PR #16, opened 2024-04-07
feat: add pixartalpha implementation

Lay2000:
The Implementation is mainly based on the existing DistriDiTPipeline.
The `prepare ` function in DistriPixArtAlphaPipeline should be modified as the archs of DiT and PixArt are different.


distrifuser/pipelines/pixartalpha.py
```diff
@@ -0,0 +1,116 @@
+import torch
```
feifeibear, line 4: 如果是copy别人代码，在开头加上出处

Lay2000: Added

PR #72, opened 2024-06-13
add patch parallelism for conv

yangpc615:
add patch parallelism for conv

### DIFF
--- a/pipefuser/modules/patch_parallel/patch_parallel_conv2d.py
+++ b/pipefuser/modules/patch_parallel/patch_parallel_conv2d.py
@@ -1,0 +1,326 @@
+# Copyright (c) 2024.6.12, PC Yang. All rights reserved.
+
+from typing import Union, Optional, Any
+from enum import Enum
+import torch
+import torch.distributed as dist
+from torch import nn
+from torch import Tensor
+from torch.nn import functional as F
+from torch.nn.common_types import _size_2_t
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class PATCH_PARA_KIND(Enum):
+    PATCH_PARA_KIND_NORMAL = 0
+    PATCH_PARA_KIND_FIRST = 1
+    PATCH_PARA_KIND_LAST = 2
+
+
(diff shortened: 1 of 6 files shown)


pipefuser/modules/patch_parallel/patch_parallel_conv2d.py
```diff
@@ -0,0 +1,337 @@
...
+            elif rank == world_size - 1:
+                padding[-2] = 0
+            else:
+                padding[-2:] = [0, 0]
+
+        if self.padding_mode != 'zeros':
+            copy_padding = self._reversed_padding_repeated_twice[:]
+            if world_size > 1:
+                adjust_padding(copy_padding, rank)
+            inner_input = F.pad(input, copy_padding, mode=self.padding_mode)
```
Eigensystem, line 143: It seems to be padding twice under the case `self.padding_mode != 'zeros' &&  self.padding != 0`. But it have no impact on the models we currently support.

Eigensystem (approved):
lgtm

pipefuser/modules/patch_parallel/patch_parallel_conv2d.py
yangpc615: There is also padding in the conv op, which I brought out because not all sides of a patch are needed to pad.  The padding parameter of  conv is set 0, such as, `F.conv2d(inner_input, weight, bias, self.stride, 0,
                            self.dilation, self.groups)`,

Eigensystem: Yeah. But I mean the duplicated F.pad in lines 143 and 161. If self.padding_mode != 'zeros' && self.padding != 0, F.pad(input, copy_padding, mode=self.padding_mode) in line 143 will be executed and then F.pad(inner_input, padding_list, mode="constant") in line 161. Is that a problem?

yangpc615: All right, I got it, and I have fixed it.

PR #77, opened 2024-06-19
feat: add patch parallelism VAE for high resolution generation

Eigensystem:
* solved vae upsampler oom problem
* Further testing and debugging required


feifeibear (approved):
LGTM

PR #86, opened 2024-06-24
fix: PatchVAE related problems

Eigensystem:
* new version of patchvae support
* fix installation script
* add switch to enable patchvae & profiler


scripts/pixart_example.py
```diff
@@ -119,6 +119,8 @@ def main():
     torch.backends.cudnn.deterministic = True
 
     enable_parallel_vae = args.use_parallel_vae
+    if args.height >= 4096:
+        enable_parallel_vae = True
```
feifeibear: we'd better let the user set it in the command line.

feifeibear (approved):

PR #90, opened 2024-06-25
Add SD3 Model

Steaunk:
Devices=8, pp=8: 1.88s
Devices=4, pp=8: 3.03s

Devices=8,pp=8
Screenshot 2024-06-25 at 3 49 27 PM


pipefuser/models/__init__.py
```diff
@@ -1,4 +1,12 @@
+from packaging.version import Version
+import diffusers
+
+if Version('0.29.0') <= Version(diffusers.__version__):
```
feifeibear, line 4: Now, SD3 needs diffusers higher than 0.29.0. However, Pixart and Dit need diffusers==0.27.2.

We should fix the dependency conflicts.

feifeibear (approved):

PR #112, opened 2024-06-27
[WIP] feat: add onediff compile to DistriTransformer2DModel, need debug

Eigensystem:
* There is a significant difference between images generated with and without using compile. The images generated with compile clearly have issues. 
* without compile
dog_pipefusion_hw_1024_sync_corrected_async_gn_sp_False_u1_w4_mb0
* with compile
dog_pipefusion_hw_1024_sync_corrected_async_gn_u0_w4_mb2_warm1

### DIFF
--- a/pipefuser/modules/dit/pipefusion/transformer_2d.py
+++ b/pipefuser/modules/dit/pipefusion/transformer_2d.py
@@ -53,6 +53,18 @@ class DistriTransformer2DModel(BaseModule):
                 start_idx:end_idx
             ]
 
+        from onediff.infer_compiler import compile
+        # options = {"dynamic": True}
+        options = {}
+        compiled_blocks = torch.nn.ModuleList(
+            [
+                # torch.compile(block, options=options, dynamic=True)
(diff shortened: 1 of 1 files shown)


pipefuser/modules/dit/pipefusion/transformer_2d.py
```diff
@@ -53,6 +53,18 @@ def __init__(self, module: Transformer2DModel, distri_config: DistriConfig):
                 start_idx:end_idx
             ]
 
+        from onediff.infer_compiler import compile
+        # options = {"dynamic": True}
+        options = {}
```
strint, line 58: Try this:
```
options = {"mode": "max-optimize:max-autotune:low-precision"}
```
or
```
options = {"mode": "max-optimize:max-autotune"}
```

Eigensystem: Thanks for your help, but it do not work. Adding "max-optimize:max-autotune" options will result in a RuntimeError: CUDA error: misaligned address.

PR #114, opened 2024-06-28
fix: warmup_timestep encoder mis-comm

Steaunk:
Currently, the difference between original image and ours is in a rational range. <del>However, there is still a bug that the 2nd image should be exactly same as the 3rd image, but there are still some subtle discrepancy.</del>

```python
query = torch.cat([query, encoder_hidden_states_query_proj], dim=1)
key = torch.cat([key, encoder_hidden_states_key_proj], dim=1)
value = torch.cat([value, encoder_hidden_states_value_proj], dim=1)
...
hidden_states, encoder_hidden_states = (
            hidden_states[:, : residual.shape[1]],
            hidden_states[:, residual.shape[1] :],
)
```
`encoder_hidden_states` depends on `hidden_states`, so the images will be different with different device number.

截屏2024-06-28 10 41 48

### DIFF
--- a/pipefuser/pipelines/pip/distri_sd3.py
+++ b/pipefuser/pipelines/pip/distri_sd3.py
@@ -298,7 +298,7 @@ class DistriSD3PiP(StableDiffusion3Pipeline):
                 else:
                     self.comm_manager.irecv_from_prev(dtype)
                     latents = self.comm_manager.get_data()
-                    if i == 0:
+                    if distri_config.rank != 1:
                         encoder_hidden_states = self.comm_manager.recv_from_prev(prompt_embeds.dtype, is_extra=True)
                 assert self._interrupt == False
                 # TBD
@@ -331,7 +331,7 @@ class DistriSD3PiP(StableDiffusion3Pipeline):
                 # if XLA_AVAILABLE:
                 #     xm.mark_step()
                 self.comm_manager.isend_to_next(latents)
-                if distri_config.rank != 0 and i == 0:
+                if distri_config.rank != 0:
                     self.comm_manager.send_to_next(next_encoder_hidden_states, is_extra=True)
 
             assert self.comm_manager.recv_queue == []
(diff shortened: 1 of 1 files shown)


Eigensystem (commented):
Has this issue been resolved, or is it still pending for the reasons mentioned above. If not resolved, could you consider doing a scatter to sync the hidden_state after the whole calculation is done. is this necessary?

Steaunk:
> Has this issue been resolved, or is it still pending for the reasons mentioned above. If not resolved, could you consider doing a scatter to sync the hidden_state after the whole calculation is done. is this necessary?

I think we can merge this PR first, current result is better than before. So far, I have no ideas about how to solve the discrepancy between different device numbers.

```python
query = torch.cat([query, encoder_hidden_states_query_proj], dim=1)
key = torch.cat([key, encoder_hidden_states_key_proj], dim=1)
value = torch.cat([value, encoder_hidden_states_value_proj], dim=1)

inner_dim = key.shape[-1]
head_dim = inner_dim // attn.heads
query = query.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)
key = key.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)
value = value.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)

hidden_states = hidden_states = F.scaled_dot_product_attention(
    query, key, value, dropout_p=0.0, is_causal=False
)
hidden_states = hidden_states.transpose(1, 2).reshape(
    batch_size, -1, attn.heads * head_dim
)
hidden_states = hidden_states.to(query.dtype)

# Split the attention outputs.
hidden_states, encoder_hidden_states = (
    hidden_states[:, : residual.shape[1]],
    hidden_states[:, residual.shape[1] :],
)
```

We split the hidden_states into different patches, and then concat each patch with encoder_hidden_states before `scaled_dot_product_attention`. I think it's not easy to get the original next encoder_hidden_states simply from combination of next encoder_hidden_states of different patches. Moreover, the next device needs the encoder_hidden_states from last device at the same time when it receives hidden_states for each patch.

Eigensystem (approved):
lgtm

PR #117, opened 2024-06-30
feat: split batch for pipefusion parallelism

Eigensystem:
* Lossless accelerated inference by splitting hidden state in batch dimension, which is 2 because of Classifier-free guidance.

## Inference speed
* pipefusion with 8 cards, 40 steps: **13.07it/s**
* pipefusion + split batch with 8 cards, splitting hidden state into 2 batch, running on 2 pipeline, each consisting of 4 cards, 40 steps: **22.08it/s**
* accelerate 1.7x on the same number of devices

## Image quality
* pipefusion parallelism with 8 cards, full pipeline.
dog_pipefusion_hw_1024_sync_corrected_async_gn_u0_w8_mb2_warm1

* pipefusion parallelism with 8 cards, split batch.
dog_pipefusion_hw_1024_sync_corrected_async_gn_u0_w8_mb2_warm1_split_batch

### DIFF
--- a/pipefuser/pipelines/pip/distri_pixartalpha.py
+++ b/pipefuser/pipelines/pip/distri_pixartalpha.py
@@ -10,7 +10,7 @@ from diffusers.pipelines.pixart_alpha.pipeline_pixart_alpha import (
 )
 from typing import List, Optional, Union, Tuple, Callable, Dict, Final
 
-import torch.distributed
+import torch.distributed as dist
 from pipefuser.utils import DistriConfig, PipelineParallelismCommManager
 from pipefuser.logger import init_logger
 
@@ -93,10 +93,10 @@ class DistriPixArtAlphaPiP(PixArtAlphaPipeline):
                 else:
                     noise_pred_uncond = torch.empty_like(noise_pred)
                     noise_pred_text = torch.empty_like(noise_pred)
-                    torch.distributed.all_gather(
+                    dist.all_gather(
                         [noise_pred_uncond, noise_pred_text],
                         noise_pred,
-                        group=self.distri_config.split_group
+                        group=self.distri_config.dp_group
                     )
                 noise_pred = noise_pred_uncond + guidance_scale * (
                     noise_pred_text - noise_pred_uncond
@@ -478,7 +478,7 @@ class DistriPixArtAlphaPiP(PixArtAlphaPipeline):
                 ):
                     progress_bar.update()
 
-            if torch.distributed.get_rank() == 0:
+            if dist.get_rank() == 0:
                 latents = torch.cat(latents, dim=2)
             else:
                 return None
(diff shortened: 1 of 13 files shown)


scripts/pixart_example.py
```diff
@@ -203,6 +211,8 @@ def main():
             num_inference_steps=args.num_inference_steps,
             output_type=args.output_type,
         )
+        if rank == 0:
+            print(210)
```
feifeibear: remove this line

pipefuser/utils.py
```diff
@@ -115,6 +115,14 @@ def __init__(
...
         self.use_resolution_binning = use_resolution_binning
 
+        self.height = height
+        self.width = width
+
+        local_rank = local_rank = int(os.getenv("LOCAL_RANK", "0"))
+        device = torch.device(f"cuda:{local_rank}")
+        torch.cuda.set_device(device)
+        self.device = device
+
```
feifeibear, line 128: name n_device_per_batch as pipefusion_degree?

pipefuser/utils.py
```diff
@@ -187,16 +191,38 @@ def __init__(self, distri_config: DistriConfig):
...
-        ) % distri_config.world_size
         self.dtype = None
-        if distri_config.world_size == 2:
-            self.groups = [torch.distributed.new_group(), torch.distributed.new_group()]
+
+
+        batch_world_size = dist.get_world_size()
+        self.batch_group = dist.new_group()
+
+        if distri_config.batch_group is not None and distri_config.split_group is not None:
```
feifeibear: 1. Use "split batch" by setting the dp degree and pipefusion degree
2. name the group for pipefusion as pipefusion_group. (split_group->pipefusion_group)
3. name the group for batch as dp_group. (batch_group -> dp_group)

Eigensystem: It feels like the DistriConfig is getting confusing as more and more options are added, is it possible to consider separating the configs for different parallel modes into different classes afterwards to make it easier to handle?

pipefuser/pipelines/pip/distri_pixartalpha.py
```diff
@@ -9,6 +9,8 @@
     retrieve_timesteps,
 )
 from typing import List, Optional, Union, Tuple, Callable, Dict, Final
+
+import torch.distributed
```
feifeibear: import torch.distributed as dist
It is not common use distributed

pipefuser/pipelines/pip/distri_pixartalpha.py
```diff
@@ -457,7 +478,7 @@ def __call__(
                 ):
                     progress_bar.update()
 
-            if distri_config.rank == 0:
+            if torch.distributed.get_rank() == 0:
```
feifeibear: not necessary

pipefuser/utils.py
```diff
@@ -166,15 +170,15 @@ def __init__(
 
     def batch_idx(self, rank: Optional[int] = None) -> int:
         if rank is None:
-            rank = self.rank
+            rank = dist.get_rank()
```
feifeibear, line 197: not necessary

pipefuser/utils.py
feifeibear: OK, this PR, we can use the split batch flag. We can change the config in the future.

feifeibear (approved):

PR #120, opened 2024-07-10
add hunyuandit model

Jenny-28:
Add hunyuandit model

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ if __name__ == "__main__":
         packages=find_packages(),
         install_requires=[
             "torch>=2.2",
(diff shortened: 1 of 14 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
         packages=find_packages(),
         install_requires=[
             "torch>=2.2",
-            "diffusers==0.29.0",
+            "diffusers==0.30.0.dev0",
```
feifeibear: Do we need a specific dev version for hunyuan?

feifeibear (approved):
LGTM

PR #128, opened 2024-07-11
Refactor: hybrid parallelism

Eigensystem:
- [x] refactor class structure
- [x] implement hybrid parallelism: DP + SP + split batch + Pipefusion


feifeibear (approved):
LGTM

Issue #137, opened 2024-07-22
Performance issue of sequence parallelism on A100 after refactor


feifeibear:
Here is my script

```
export SCRIPT=pixartalpha_example.py
# export SCRIPT=sd3_example.py
export MODEL_ID="/mnt/models/SD/PixArt-XL-2-1024-MS"
# export MODEL_ID="/mnt/models/SD/HunyuanDiT-v1.2-Diffusers"
# export MODEL_ID="/mnt/models/SD/stable-diffusion-3-medium-diffusers"


for HEIGHT in 2048
do
for N_GPUS in 8;
do 


torchrun --nproc_per_node=$N_GPUS \
./examples/$SCRIPT \
--model $MODEL_ID \
--pipefusion_parallel_degree 1 \
--ulysses_degree 8 \
--ring_degree 1 \
--num_inference_steps 20 \
--warmup_steps 0 \
--height $HEIGHT \
--width $HEIGHT \
--num_pipeline_patch 8 \
--no_use_resolution_binning \
--prompt "A small dog" \


done
done
```

1. pipefusion_parallel_degree=8, ulysses_degree=1, ring_degree=1, not  use_split batch, num_pipeline_patch=8
epoch time: 3.98 sec, memory: 22.112337408 GB
legacy：3.37 sec

2. pipefusion_parallel_degree=1, ulysses_degree=8, ring_degree=1, not  use_split batch, num_pipeline_patch=8
epoch time: 11.00 sec, memory: 15.671771136 GB
legacy：2.46

3. pipefusion_parallel_degree=1, ulysses_degree=8, ring_degree=1, not  use_split batch, num_pipeline_patch=8
epoch time: 13.20 sec, memory: 15.671771136 GB

4. pipefusion_parallel_degree=1, ulysses_degree=4, ring_degree=1,  use_split batch, num_pipeline_patch=8
epoch time: 25.48 sec, memory: 23.379786752 GB

5. pipefusion_parallel_degree=4, ulysses_degree=1, ring_degree=1,  use_split batch, num_pipeline_patch=8
epoch time: 3.57 sec, memory: 22.901449728 GB

@Eigensystem told me that num_pipeline_patch should be 1. still very slow.
6. pipefusion_parallel_degree=4, ulysses_degree=1, ring_degree=1,  use_split batch, num_pipeline_patch=1
epoch time: 25.48 sec, memory: 23.379786752 GB

# problems
[...]

PR #142, opened 2024-07-22
Fix: sequence parallel performance

Eigensystem:


Eigensystem:
* sp_degree == 8 with legacy version: 6.10s, hybrid parallelism version: 7.72s
* sp_degree == 4 + split batch with hybrid parallelism version: 2.70s
* sp_degree == 2 + pp_degree == 2 + split_batch: 1.05s

Tested on 8 * L40

Due to some conflicts in the patch segmentation methods between sequence parallel and pipeline fusion parallel, if patches are divided according to the existing method, it becomes impossible to fully utilize the advantages of SP and PP in a hybrid parallel scenario. We could try to optimize the division method to achieve optimal performance in mixed parallel situations.

#137

Eigensystem:
* the patch processing now:
image

feifeibear (approved):
LGTM

PR #144, opened 2024-07-24
[WIP] Adding SD3 into xDiT

dannyxiaocn:
Using the new developing interface to implement Stable Diffusion 3
Still fixing bugs on parallelisms and version problem


pipefuser/models/transformers/transformer_sd3.py
```diff
@@ -0,0 +1,165 @@
...
+            scale_lora_layers(self, lora_scale)
+        else:
+            if joint_attention_kwargs is not None and joint_attention_kwargs.get("scale", None) is not None:
+                logger.warning(
+                    "Passing `scale` via `joint_attention_kwargs` when not using the PEFT backend is ineffective."
+                )
+
+        height, width = hidden_states.shape[-2:]
+
+        hidden_states = self.pos_embed(hidden_states)  # takes care of adding positional embeddings too.
```
feifeibear: should we do pos embed on pp_rank=0?

dannyxiaocn: Thank you, I have updated this part

PR #159, opened 2024-08-02
refactor:add is_pipeline_first_stage,is_pipeline_last_stage and so on

taozhiwei:
1. add fun: is_pipeline_first_stage,is_pipeline_last_stage;
2. always use all_gather_into_tensor replace all_gather,because in the implementation of torch, the list will be flattened and unflattened, which will result in additional allocation of GPU memory and D2D operations.;
3. when not last pipeline stage set norm_out and proj_out to none;


Steaunk:
Are there any tests about the time & output pictures?

examples/pixartalpha_example.py
```diff
@@ -1,22 +1,18 @@
...
 import torch.distributed
 from xfuser import xFuserPixArtAlphaPipeline, xFuserArgs
 from xfuser.config import FlexibleArgumentParser
-from xfuser.distributed import (
-    get_world_group, 
-    get_data_parallel_rank, 
-    get_data_parallel_world_size,
-    get_runtime_state
-)
+from xfuser.distributed import ps, rs
```
Eigensystem, line 7: I don't think it is a good idea to use representations like `ps` and `rs` as they add code with little benefit. It seems to me that it would be more straightforward and convenient to use the entire distributed directory for runtime environment management and import the required functions directly from xfuser.distributed

taozhiwei: > I don't think it is a good idea to use representations like `ps` and `rs` as they add code with little benefit. It seems to me that it would be more straightforward and convenient to use the entire distributed directory for runtime environment management and import the required functions directly from xfuser.distributed

使用ps和rs后，可以避免每次需要用某个函数时，还需专门再去导入一下；我是借鉴了别的框架的用法，我看Megatron是用mpu，DeepSpeed是用groups来管理这些类似函数的。

taozhiwei:
> Are there any tests about the time & output pictures?



> Are there any tests about the time & output pictures?

图片生成都是正常的，我在8卡V100上，加上--use_split_batch测试，跑pixartalpha_example.py模型；5次均值由4.448降为4.435

examples/pixartalpha_example.py
Eigensystem: Maybe a statement like `import xfuser.distributed as (a name like mpu)` (or rename distributed dir) could be used to solve the problem? It doesn't feel like it's necessary to distinguish between ps and rs, respectively.

feifeibear: 我认为没有必要用ps和rs，并没有给使用带来便利，而且ps和rs这两个名称非常confusing

taozhiwei: > 我认为没有必要用ps和rs，并没有给使用带来便利，而且ps和rs这两个名称非常confusing

使用上的便利还是有的，第一就是需要用某个函数时，不需要显示导入了，第二就是编码时，更利于智能提示，比如在vscode里面，输入ps.,就会把候选函数列出来。这2个简写名称可能需要改一下？

taozhiwei:
https://github.com/xdit-project/xDiT/pull/166  
重新提交了一个, @Eigensystem @feifeibear

PR #166, opened 2024-08-06
refactor:add is_pipeline_first_stage,is_pipeline_last_stage and so on

taozhiwei:
1. add fun: is_pipeline_first_stage,is_pipeline_last_stage，is_dp_last_rank;
2. always use all_gather_into_tensor replace all_gather,because in the implementation of torch, the list will be flattened and unflattened, which will result in additional allocation of GPU memory and D2D operations.;
3. when not last pipeline stage set norm_out and proj_out to none;
4. fix bug，When there is no results directory；


xfuser/distributed/group_coordinator.py
```diff
@@ -189,9 +189,9 @@ def all_reduce(self, input_: torch.Tensor) -> torch.Tensor:
         return input_
 
     def all_gather(
-        self, 
-        input_: torch.Tensor, 
-        dim: int = -1, 
+        self,
+        input_: torch.Tensor,
+        dim: int = 0,
```
feifeibear, line 194: 为什么更改了默认值？

taozhiwei: > 为什么更改了默认值？

因为allgather默认是在dim 0做的，这个dim -1是最后一个维度，原来separate_tensors为true的时候，没有考虑dim这个参数。现在为了统一起来，就改成0了。

feifeibear (approved):
LGTM

PR #174, opened 2024-08-08
feat: add parallel condition checking decorator

Eigensystem:
- add parallel condition checking before `__call__`
- move sync & async pipeline to implementation class instead of base class


feifeibear (approved):
LGTM

PR #203, opened 2024-08-20
add feedforward TP to Flux 1.0

feifeibear:


xfuser/model_executor/layers/feedforward.py
```diff
@@ -1,25 +1,23 @@
...
 
 @xFuserLayerWrappersRegister.register(FeedForward)
 class xFuserFeedForwardWrapper(xFuserLayerBaseWrapper):
-    def __init__(self, module: FeedForward):
-        super(xFuserFeedForwardWrapper, self).__init__(module=module)
+    def __init__(self, feedforward: FeedForward):
+        super(xFuserFeedForwardWrapper, self).__init__(module=feedforward)
 
-        self.module = module
+        self.module = feedforward
```
Eigensystem: not necessary

feifeibear: DONE

PR #259, opened 2024-09-09
Refactor setup.py and add instruction for cu124

lixiang007666:
This PR is done:

- [x] Refactored setup.py, with special attention to adding clear installation instructions for users with CUDA 12.4 to prevent errors with `torchrun`.


Eigensystem (approved):
LGTM.

PR #280, opened 2024-09-20
fix: use xFuserJointLongContextAttention as the underlying SP method for CogVideoX

xibosun:
This pr makes the following changes:

1. Within the `xFuserCogVideoXAttnProcessor2_0` class, replace `xFuserLongContextAttention` with `xFuserJointLongContextAttention` to perform SP execution correctly.
2. Fix the `forward` function in the `xFuserJointLongContextAttention` class to support various `joint_strategies`.


xfuser/model_executor/layers/attention_processor.py
```diff
@@ -121,7 +121,7 @@ def __init__(
         assert (to_k.bias is None) == (to_v.bias is None)
         assert to_k.weight.shape == to_v.weight.shape
 
-        in_size, out_size = to_k.in_features, to_k.out_features
+        '''in_size, out_size = to_k.in_features, to_k.out_features
```
feifeibear: use # for comments or release these lines.

feifeibear (changes requested):

xfuser/model_executor/layers/attention_processor.py
xibosun: fixed.

feifeibear (approved):

Issue #271, opened 2024-09-19
Parallel VAE fails when the degree of parallelism > 1


gty111:
When generating images with `--use_parallel_vae`, the program hangs. Because only the last rank of PP or SP execute `self.vae.decode` , while all the ranks are expected.

https://github.com/xdit-project/xDiT/blob/e506417e3c4bb6cf18957566bd86ef3755d3229f/xfuser/model_executor/pipelines/pipeline_stable_diffusion_3.py#L382-L392

PR #281, opened 2024-09-21
Fix parallel vae

gty111:
related to #271

### DIFF
--- a/xfuser/model_executor/pipelines/pipeline_flux.py
+++ b/xfuser/model_executor/pipelines/pipeline_flux.py
@@ -32,6 +32,7 @@ from xfuser.core.distributed import (
     is_pipeline_first_stage,
     is_pipeline_last_stage,
     is_dp_last_group,
+    get_world_group
 )
 from .base_pipeline import xFuserPipelineBaseWrapper
 from .register import xFuserPipelineWrapperRegister
@@ -309,19 +310,30 @@ class xFuserFluxPipeline(xFuserPipelineBaseWrapper):
                     sync_only=True,
                 )
 
+        def vae_decode(latents):
+            latents = self._unpack_latents(
+                    latents, height, width, self.vae_scale_factor
+            )
+            latents = (
+                latents / self.vae.config.scaling_factor
+            ) + self.vae.config.shift_factor
+            
+            image = self.vae.decode(latents, return_dict=False)[0]
+            return image
+        
+        if not output_type == "latent":
+            if get_runtime_state().runtime_config.use_parallel_vae:
+                latents = get_world_group().broadcast_latent(latents,get_runtime_state().runtime_config.dtype)
+                image = vae_decode(latents)
+            else:
+                if is_dp_last_group():
+                    image = vae_decode(latents)
+        
         if is_dp_last_group():
             if output_type == "latent":
                 image = latents
 
             else:
-                latents = self._unpack_latents(
-                    latents, height, width, self.vae_scale_factor
-                )
-                latents = (
-                    latents / self.vae.config.scaling_factor
-                ) + self.vae.config.shift_factor
-
-                image = self.vae.decode(latents, return_dict=False)[0]
                 image = self.image_processor.postprocess(image, output_type=output_type)
 
             # Offload all models
--- a/xfuser/model_executor/pipelines/pipeline_flux.py
+++ b/xfuser/model_executor/pipelines/pipeline_flux.py
@@ -323,13 +323,13 @@ class xFuserFluxPipeline(xFuserPipelineBaseWrapper):
         
         if not output_type == "latent":
             if get_runtime_state().runtime_config.use_parallel_vae:
-                latents = get_world_group().broadcast_latent(latents,get_runtime_state().runtime_config.dtype)
+                latents = self.gather_broadcast_latents(latents)
                 image = vae_decode(latents)
             else:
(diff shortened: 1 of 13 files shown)


gty111:
In this PR, also fix #282

xfuser/model_executor/models/transformers/transformer_flux.py
```diff
@@ -109,21 +109,20 @@ def forward(
...
-            txt_ids = txt_ids[0]
-        if img_ids.ndim == 3:
-            logger.warning(
-                "Passing `img_ids` 3d torch.Tensor is deprecated."
-                "Please remove the batch dimension and pass it as a 2d torch Tensor"
-            )
-            img_ids = img_ids[0]
-
-        ids = torch.cat((txt_ids, img_ids), dim=0)
+        # if txt_ids.ndim == 3:
```
Eigensystem: why delete the check

xfuser/model_executor/pipelines/pipeline_flux.py
```diff
@@ -309,19 +310,30 @@ def __call__(
...
+            latents = (
+                latents / self.vae.config.scaling_factor
+            ) + self.vae.config.shift_factor
+            
+            image = self.vae.decode(latents, return_dict=False)[0]
+            return image
+        
+        if not output_type == "latent":
+            if get_runtime_state().runtime_config.use_parallel_vae:
+                latents = get_world_group().broadcast_latent(latents,get_runtime_state().runtime_config.dtype)
```
Eigensystem: It's not very rigorous. If we enable dp/cfg，we can not broadcast like this.

xfuser/model_executor/pipelines/pipeline_pixart_sigma.py
```diff
@@ -323,11 +324,23 @@ def __call__(
...
+        
+        def vae_decode(latents):
+            image = self.vae.decode(
+                latents / self.vae.config.scaling_factor, return_dict=False
+            )[0]
+            return image
+        
+        if not output_type == "latent":
+            if get_runtime_state().runtime_config.use_parallel_vae:
+                latents = get_world_group().broadcast_latent(latents,get_runtime_state().runtime_config.dtype)
```
Eigensystem: as above

xfuser/model_executor/pipelines/pipeline_stable_diffusion_3.py
```diff
@@ -379,16 +380,26 @@ def __call__(
...
+        def vae_decode(latents):
+            latents = (
+                    latents / self.vae.config.scaling_factor
+                ) + self.vae.config.shift_factor
+            image = self.vae.decode(latents, return_dict=False)[0]
+            return image
+        
+        if not output_type == "latent":
+            if get_runtime_state().runtime_config.use_parallel_vae:
+                latents = get_world_group().broadcast_latent(latents,get_runtime_state().runtime_config.dtype)
```
Eigensystem: as above

Eigensystem (changes requested):

xfuser/model_executor/models/transformers/transformer_flux.py
gty111: In diffuser v0.30.3, it supports passing three dimensions tensor. So we do not need to lower the dimension of the tensor.
https://github.com/huggingface/diffusers/blob/v0.30.3/src/diffusers/models/transformers/transformer_flux.py#L380

xfuser/model_executor/pipelines/pipeline_flux.py
gty111: In the following commit, fix this issue.

gty111:
To consider DP, we add `gather_broadcast_latents` and `is_dp_last_group` in `xFuserPipelineBaseWrapper`. 
`gather_broadcast_latents`: 
1. identify dp last group 
2. gather latents from dp last group and concatenate the latents in the batch dim
3. broadcast latents 

Then use default group to complete parallel vae.

When using parallel vae, only one process need to process img after vae. So `is_dp_last_group` is to correctly identify the process which needs to do sth after vae.

@Eigensystem

xfuser/model_executor/pipelines/pipeline_hunyuandit.py
```diff
@@ -454,12 +455,22 @@ def __call__(
...
+            )[0]
+            return image
+        
+        if not output_type == "latent":
+            if get_runtime_state().runtime_config.use_parallel_vae:
+                latents = self.gather_broadcast_latents(latents)
+                vae_decode(latents)
+            else:
+                if is_dp_last_group():
+                    vae_decode(latents)
```
Eigensystem: miss return value of func `vae_decode`

Eigensystem (changes requested):

Eigensystem:
I think it is not a elegant implementation because you combined all the dp latents into one tensor instead of using different dp groups to process different latent. It will increase the communication load.

gty111:
> I think it is not a elegant implementation because you combined all the dp latents into one tensor instead of using different dp groups to process different latent. It will increase the communication load.

It is indeed not the best implementation. Using different dp groups need to refactor DistVAE. Since parallel vae is an option, maybe we can first fix parallel vae in this PR, and further optimize it in the future.

Eigensystem (approved):
LGTM.

PR #297, opened 2024-10-08
feat: DiTFastAttn for PixArt

ZDJeffrey:
## Summary
[DiTFastAttn](https://github.com/thu-nics/DiTFastAttn) is an attention compression method for Diffusion Transformer Models. Using the redundancy of DiT. It introduces some compression methods for the self-attention to accelerate the inference speed on a single GPU. 

## Implementation
- A new config class and a new argument group are added for the DiTFastAttn in `xfuser\config\`
- Following the implementation of Long Context Attention, DiTFastAttn module is implemented in `xfuser\core\fast_attention`
- As it can only be used with data parallelism, the attention processor is implemented independently instead of greatly modify the original attention processor in `xfuser\model_executor\layers\attention_processor.py`.
- Before the DiTFastAttn works, the compression method need to be set. Thus when the function `prepare_run` is called, the compression method is set if the DiTFastAttn is enabled.

## How to use
- To use the DiTFastAttn, the following arguments are required:
  - `use_fast_attn`: enable the fast attention.
  - `n_calib`: Number of prompts for compression method selection.
  - `threshold`: Threshold for selecting attention compression method. It relatively determines the compression ratio.
  - `window_size`: Size of window attention. According to the paper, the window size is recommended to be 1/8 of the token size.
- When using the DiTFastAttn for a model for the first time, some arguments need to be set for the compression.
  - `coco_path`: Path of MS COCO annotation json file(e.g. `captions_val2014.json` from [official site](https://cocodataset.org/#download)). The file contains the captions of the images, which are sampled for the compression.
- After the compression, the method will be saved in a json file in `cache` folder. This file can be loaded for the same models with same arguments if `use_cache` is set.
- Note: The DiTFastAttn can only be used with data parallelism. If other parallelism methods are used, the program will raise an error.

## Test
So far, only the implementation of DiTFastAttn for PixArt models are done. I have tested it with data parallelism on `PixArt-alpha/PixArt-Sigma-XL-2-1024-MS` and `PixArt-alpha/PixArt-XL-2-1024-MS`. As the `PixArt-alpha/PixArt-Sigma-XL-2-2K-MS` model is not available on the Huggingface now, I have not tested it yet.

## WIP
Implementation of DiTFastAttn for other models is still in progress. 
The benchmark of the DiTFastAttn is not done yet. I will do it after the implementation of DiTFastAttn for other models is done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ xfuser.egg-info/
 dist/*
 latte_output.mp4
 *.sh
+cache/
--- a/examples/run.sh
+++ b/examples/run.sh
@@ -19,7 +19,7 @@ export PYTHONPATH=$PWD:$PYTHONPATH
 # or you can simply use the model's ID on Hugging Face, 
 # which will then be downloaded to the default cache path on Hugging Face.
 
-export MODEL_TYPE="CogVideoX"
+export MODEL_TYPE="Pixart-alpha"
 # Configuration for different model types
 # script, model_id, inference_step
 declare -A MODEL_CONFIGS=(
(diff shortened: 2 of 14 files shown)


ZDJeffrey:
## Image Compare
### PixArt-alpha/PixArt-XL-2-1024-MS
- origin
  - origin1
  - origin2
- DiTFastAttn(threshold=0.15)
  - fastattn1
  - fastattn2

Eigensystem (approved):
LGTM. Elegant code!

xibosun (approved):
The pr defines FastAttnState, with a similar implementation to the existing state. Most code of DiTFastAttn are written to xfuser/core/fast_attention with minor modification to existing code. Overall, the code is correct and elegant.

asahni04:
how to use DiTFastAttn generically inplace/with of say just Flash Attention-2 or Flash attention-3. say i have the following like to compute attention:

```
from flash_attn_interface import flash_attn_func
some norms applied to q,k,v
q
k
v
 x = flash_attn_func(q, k, v)[0]
 x = rearrange(x, "b n h d -> b n (h d)")
 x = self.proj(x)
```

PR #327, opened 2024-10-30
support FID compute

Lay2000:
Support FID comput


xibosun (approved):
The FID scripts are correct.

PR #376, opened 2024-12-03
Optimize USP interface for Flux and CogVideo

xibosun:


xfuser/model_executor/layers/usp_pre250.py
```diff
@@ -0,0 +1,105 @@
+# This file implements USP with torch version < '2.5.0'
```
feifeibear: change the file name to usp_legacy?

xibosun: sure.

docs/performance/cogvideo.md
```diff
@@ -43,3 +43,17 @@ As shown in the figure, regardless of Ulysses Attention, Ring Attention or CFG p
     <img src="https://raw.githubusercontent.com/xdit-project/xdit_assets/main/performance/cogvideo/latency-cogvideo1.5-5b-l40.png" 
     alt="latency-cogvideo1.5-5b-l40">
 </div>
+
+xDiT can also achieve similar acceleration effects when generating 81-frame 1360x768 resolution video on H20 and L20.
```
feifeibear: give details of latency here. and compare the ROI of two GPUs, which is better?

feifeibear (approved):

PR #401, opened 2024-12-19
add hunyuan_video_usp_example.py

chengzeyi:


feifeibear (approved):
LGTM

examples/hunyuan_video_usp_example.py
```diff
@@ -0,0 +1,303 @@
...
+
+    new_forward = new_forward.__get__(transformer)
+    transformer.forward = new_forward
+
+    for block in transformer.transformer_blocks + transformer.single_transformer_blocks:
+        block.attn.processor = xFuserHunyuanVideoAttnProcessor2_0()
+
+
+def main():
+    parser = FlexibleArgumentParser(description="xFuser Arguments")
```
feifeibear, line 209: I suggest check diffusers version here.

chengzeyi: still only works on nightly diffusers, will add it when it becomes stable~

feifeibear (approved):

PR #426, opened 2025-01-08
[Fix] Add method to prepare extra step kwargs for scheduler in xFuserCogVideoXPipeline

LazyBusyYang:
### What's the problem
`self.scheduler` is an instance of `xFuserDDIMSchedulerWrapper`, whose `step` is defined like `def step(self, *args, **kwargs)`. The original `accepts_eta` and `accepts_generator` are always false because `set(inspect.signature(self.scheduler.step).parameters.keys())` returns a set of args and kwargs.

This update introduces the `prepare_extra_step_kwargs` method, which prepares additional keyword arguments for the scheduler step based on its signature. It checks for the presence of 'eta' and 'generator' parameters to ensure compatibility with different schedulers. This enhancement improves the flexibility and usability of the pipeline.

### DIFF
--- a/xfuser/model_executor/pipelines/pipeline_cogvideox.py
+++ b/xfuser/model_executor/pipelines/pipeline_cogvideox.py
@@ -3,6 +3,7 @@ from typing import Any, List, Tuple, Callable, Optional, Union, Dict
 
 import torch
 import torch.distributed
+import inspect
 from diffusers import CogVideoXPipeline
 from diffusers.pipelines.cogvideo.pipeline_cogvideox import (
     CogVideoXPipelineOutput,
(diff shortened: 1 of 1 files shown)


xfuser/model_executor/pipelines/pipeline_cogvideox.py
```diff
@@ -404,6 +405,24 @@ def _init_sync_pipeline(
             )
         return latents, prompt_embeds, image_rotary_emb
 
+
+    def prepare_extra_step_kwargs(self, generator, eta):
```
feifeibear, line 409: where did you call this function?

LazyBusyYang: Here [extra_step_kwargs = self.prepare_extra_step_kwargs(generator, eta)](https://github.com/xdit-project/xDiT/blob/ec182a32ce587435cc1217f701ab3b2a5d1cd1ff/xfuser/model_executor/pipelines/pipeline_cogvideox.py#L248C8-L248C75). In my case, `print(extra_step_kwargs)` on main branch always gives me an empty dict.

PR #434, opened 2025-01-16
[Bugfix] avoid yunchang undefined & format

DefTruth:


xfuser/model_executor/layers/attention_processor.py
```diff
@@ -195,7 +197,9 @@ def __init__(self):
                     use_kv_cache=self.use_long_ctx_attn_kvcache
                 )
             else:
+                import yunchang
```
DefTruth, line 200: add import yunchang here

feifeibear (approved):
LGTM

PR #438, opened 2025-01-17
[Tests]: Context and Sequence Parallel 

antferdom:
Guarantee distributed inference integrity with newer `xFuser` versions (e.g. [world_size -> dit_parallel_size](https://github.com/xdit-project/xDiT/blob/b2bcc38a23b10abcd14c46346fabff50b9fc7d33/xfuser/config/config.py#L112)) and gather benchmarking results. [ParaAttention testing is our main reference](https://github.com/chengzeyi/ParaAttention/blob/main/tests/context_parallel/test_diffusers_adapters.py).


antferdom:
@feifeibear @xibosun

tests/context_parallel/debug_tests.py
```diff
@@ -0,0 +1,18 @@
+import os
+import sys
```
feifeibear, line 2: run_test_diffusers_adapter.py

http-service/load_test.sh
```diff
@@ -0,0 +1,6 @@
+HOST=127.0.0.1
```
feifeibear, line 1: change name to run_test_clinet.sh?

tests/context_parallel/debug_flux_usp_example.py
```diff
@@ -0,0 +1,44 @@
+import os
```
feifeibear, line 1: launch_flux_usp_service_example.py?

feifeibear (changes requested):
Thanks for the MR! I suggest to change names of some files.

tests/context_parallel/debug_flux_usp_example.py
antferdom: Sure, these were temporal scripts meant for debugging, thus we can directly remove it. Or we can keep it as a blueprint for debugging.

antferdom:
> Thanks for the MR! I suggest to change names of some files.

Before merging I wanted to have feedback about adding tests covering all the models supported by xDiT. What do you think?

feifeibear:
> > Thanks for the MR! I suggest to change names of some files.
> 
> Before merging I wanted to have feedback about adding tests covering all the models supported by xDiT. What do you think?

Thank you for your thoughtful consideration. I certainly hope you can cover as many model services as possible.

PR #443, opened 2025-01-23
[Bugfix] Fix for only one DIT worker and a separate VAE worker.

lihuahua123:
When VAE_PARALLEL_SIZE=1 and DIT_PARALLEL_SIZE>=1, it does not use use_naive_forward.


feifeibear (approved):
LGTM

PR #480, opened 2025-03-19
fix rlt when using fb cache for flux

yinfan98:
Adapt from: https://github.com/chengzeyi/ParaAttention/blob/main/doc/fastest_flux.md
I find xDiT use rlt to 0.6, This will cause a degradation in generate quality. 0.12 rlt might be a more reasonable value.

When using rlt to 0.6 may cause: 
WechatIMG746

Compare with 0.12 and not cache: 
WechatIMG747


yinfan98:
prompt: A small dog

feifeibear (approved):

PR #504, opened 2025-04-19
Update get_2d_sincos_pos_embed to use output_type='pt' for Diffusers compat

fielding:
**Description**  
This PR addresses a deprecation warning in Diffusers 0.33.0 related to the `output_type` parameter in `get_2d_sincos_pos_embed`. The function now recommends using `output_type='pt'` to return a PyTorch tensor directly, which is more efficient and avoids unnecessary conversions from NumPy.

**Changes**  
- Modified `get_2d_sincos_pos_embed` in `xfuser/model_executor/layers/embeddings.py` to use `output_type='pt'`.  
- Added the `device` parameter to ensure the tensor is placed directly on the correct device, removing the need for manual device transfer.  
- Removed the previous `torch.from_numpy` conversion and `.to(latent.device)` call, as they are no longer required these changes.

**Benefits**  
- Ensures xFuser remains compatible with Diffusers' updated API, avoiding future issues.  
- Improves performance by eliminating NumPy-to-PyTorch conversions and redundant device transfers.  
- Simplifies the code.

**Testing**  
- Tested with Diffusers 0.33.0.dev0 live on a 8x H200 machine. Can report back testings on a more variety of machines if needed.

Cheers

### DIFF
--- a/xfuser/model_executor/layers/embeddings.py
+++ b/xfuser/model_executor/layers/embeddings.py
@@ -71,9 +71,10 @@ class xFuserPatchEmbedWrapper(xFuserLayerBaseWrapper):
                     grid_size=(height, width),
                     base_size=self.module.base_size,
                     interpolation_scale=self.module.interpolation_scale,
+                    device=latent.device,  # Place tensor on the correct device directly
+                    output_type="pt",  # Return a PyTorch tensor instead of NumPy array
                 )
-                pos_embed = torch.from_numpy(pos_embed)
-                self.module.pos_embed = pos_embed.float().unsqueeze(0).to(latent.device)
+                self.module.pos_embed = pos_embed.unsqueeze(0)  # Shape adjustment
                 self.module.height = height
                 self.module.width = width
                 pos_embed = self.module.pos_embed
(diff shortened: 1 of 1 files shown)


xfuser/model_executor/layers/embeddings.py
```diff
@@ -71,9 +71,10 @@ def forward(self, latent):
                     grid_size=(height, width),
                     base_size=self.module.base_size,
                     interpolation_scale=self.module.interpolation_scale,
+                    device=latent.device,  # Place tensor on the correct device directly
```
feifeibear: Does this change work with lower versions of diffusers, such as 0.32.0? If it's only compatible with 0.33.0, we could implement an if-else statement with a diffuser version check as the condition."

fielding: Ooops, ignore the vanishing comments.

Good catch! I was in a bit of a hurry and didn't check when the arguments were added, haha. I confirmed it was 0.33.0, so I've updated it to use the simplified code conditionally. 

Appreciate you pointing that out! Let me know if anything else is needed! Cheers!

feifeibear (approved):

PR #505, opened 2025-04-21
fix: support use_sync flag in xFuserLongContextAttention

Lay2000:


feifeibear (approved):

Ethkuil:
It would be better if comments are updated too. Certainly this doesnt matter though.

PR #513, opened 2025-05-14
fix: keep guidance_scale default value consistent with diffusers

ZDJeffrey:
Since the default value of guidance_scale in diffusers was changed from 7.0 to 3.5, the outputs of examples/flux_examples.py now differ between the XDiT forward and the naive forward implementations.
- default value of guidance_scale in diffusers
https://github.com/huggingface/diffusers/blob/f4fa3beee7f49b80ce7a58f9c8002f43299175c9/src/diffusers/pipelines/flux/pipeline_flux.py#L640
- diffusers commit
https://github.com/huggingface/diffusers/commit/089cf798eb199ddc0d396c7bbb0172fecf2845e9

### DIFF
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -100,7 +100,7 @@ class xFuserArgs:
     no_use_resolution_binning: bool = False
     seed: int = 42
     output_type: str = "pil"
-    guidance_scale: float = 0.0
+    guidance_scale: float = 3.5
     enable_model_cpu_offload: bool = False
     enable_sequential_cpu_offload: bool = False
     enable_tiling: bool = False
@@ -294,7 +294,7 @@ class xFuserArgs:
         input_group.add_argument(
             "--guidance_scale",
             type=float,
-            default=0.0,
+            default=3.5,
             help="Guidance scale for classifier free guidance.",
         )
         runtime_group.add_argument(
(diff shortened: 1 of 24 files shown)


feifeibear:
Does it work for the low version diffusers? 

We'd better set the guild scale in example.py.

ZDJeffrey:
> Does it work for the low version diffusers?
> 
> We'd better set the guild scale in example.py.

Okay, I've added the corresponding argument and specified it in the example.sh script.

xfuser/config/config.py
```diff
@@ -263,6 +263,7 @@ class InputConfig:
     max_sequence_length: int = 256
     seed: int = 42
     output_type: str = "pil"
+    guidance_scale: float = 0.0
```
feifeibear: set it default as 3.5

xfuser/config/args.py
```diff
@@ -290,6 +291,12 @@ def add_cli_args(parser: FlexibleArgumentParser):
             default="pil",
             help="Output type of the pipeline.",
         )
+        input_group.add_argument(
+            "--guidance_scale",
+            type=float,
+            default=0.0,
```
feifeibear: default as 3.5

xfuser/config/args.py
```diff
@@ -100,6 +100,7 @@ class xFuserArgs:
     no_use_resolution_binning: bool = False
     seed: int = 42
     output_type: str = "pil"
+    guidance_scale: float = 0.0
```
feifeibear: default as 3.5

feifeibear (approved):

PR #522, opened 2025-05-29
ADD FP8 forward for FA3

yuyu5333:
FA3 now supports FP8 (torch.float8_e4m3fn) reasoning. However, the forward stage still lacks FP8 support. 

I added the necessary parameters qkv scale for quantization and selected the Attention that supports FP8 for forwarding according to AttnType in yunchang (USP). This function will support USP forward of FP8 quantization.

In addition, pull requests have been initiated for quantization-related parameter support for yunchang and FA3.

yunchang (USP): https://github.com/feifeibear/long-context-attention/pull/151（merged into main）
FA3: https://github.com/Dao-AILab/flash-attention/pull/1686


feifeibear (approved):
LGTM

PR #539, opened 2025-07-18
[feature] adapt for Moore Threads graphics processing unit

houchen-li:
# Description

Adding Moore Threads GPU support, expanding xDiT compatibility with MUSA—Moore Threads’ software stack for AI and graphics acceleration on MT chips.

# Features

* add support for Moore Threads GPU Family (MTT-S80, MTT-S4000, etc.)

# Risks

*  This patch won't interfere CUDA or HIP setup if your system does not include MUSA software.

PR #540, opened 2025-07-21
[fix] move all platform dependent code to xfuser/envs.py

houchen-li:
# Description

The source code of xDiT contains few problems:

1. In file `xfuser/core/long_ctx_attention/ring/ring_flash_attn.py:5` , `flex_attention` was imported but never be used. Since this module is introduced in PyTorch 2.5， adding this unused module may prevent xDiT from applying to the models which requires PyTorch < 2.5.

2. In file `xfuser/core/long_ctx_attention/ring/ring_flash_attn.py:108` , the implicit default `softmax_scale` was calculated via `q.shape[-1] ** (-0.5)`, which truncates to an integer number that brokes the entire workflow. The right way to calculate this value is by `1.0 / math.sqrt(q.size(-1))` which ends up with a float-point number.

3. In `xfuser/core/distributed/parallel_state.py`, [previous patch](https://github.com/xdit-project/xDiT/pull/539) involves platform-dependent codes with try-except blocks. This can be improved by moving all platform-dependent codes to `xfuser/envs.py`.

# Features

* remove unused module `flex_attention` from `xfuser/core/long_ctx_attention/ring/ring_flash_attn.py`.
* get rid of the type trunction error when calculate implicit default `softmax_scale`.
* move platform dependent codes to `xfuser/envs.py`

# Risks

* xDiT is now available on PyTorch < 2.5 environment. But I haven't test lots of those models. The stability and accuracy may not be guaranteed.

### DIFF
--- a/xfuser/core/distributed/group_coordinator.py
+++ b/xfuser/core/distributed/group_coordinator.py
@@ -136,14 +136,7 @@ class GroupCoordinator:
         assert self.cpu_group is not None
         assert self.device_group is not None
 
-        if torch.cuda.is_available():
-            self.device = torch.device(f"cuda:{local_rank}")
-        else:
-            try:
-                if torch.musa.is_available():
-                    self.device = torch.device(f"musa:{local_rank}")
-            except ModuleNotFoundError:
-                self.device = torch.device("cpu")
+        self.device = envs.get_device(local_rank)
 
     @property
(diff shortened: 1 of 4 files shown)


houchen-li:
@feifeibear

feifeibear (approved):

PR #559, opened 2025-09-09
feat: Add Apple Silicon (MPS) support

haqatak:
This change adds support for Apple Silicon, allowing the library to run on MPS-compatible devices.

Key changes:
- Modified `xfuser/envs.py` to detect and support the MPS backend.
- Replaced hardcoded `.cuda()` calls with dynamic device placement (`.to(device)`).
- Made CUDA-specific dependencies like `yunchang` and `flash-attn` conditional, preventing import errors on non-CUDA systems.
- Updated `setup.py` to be more platform-agnostic.
- Added a new test file `tests/core/test_envs.py` to verify the new device-detection logic.

Limitation:
The `yunchang` library for long context attention is CUDA-specific and is therefore disabled when running on MPS. This means that long context attention features will not be available on Apple Silicon devices.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ def get_cuda_version():
         ]
         cuda_version = version_line.split(" ")[-2].replace(",", "")
         return "cu" + cuda_version.replace(".", "")
-    except Exception as e:
+    except (subprocess.CalledProcessError, FileNotFoundError):
         return "no_cuda"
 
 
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@ def get_cuda_version():
         author_email="fangjiarui123@gmail.com",
         packages=find_packages(),
         install_requires=[
-            "torch>=2.1.0",
+            "torch==2.4.1",
```
feifeibear, line 29: can you make it as torch>=2.4.1

feifeibear (approved):

PR #571, opened 2025-09-29
Refactor packages info to remove conflicting logging

avjves:
# What
Small refactoring to gate missing `flash_attn` warning logs behind not having `AITER` installed.

# Why
Currently when running xDiT on AMD devices, one can use AITER instead of `flash_attn`, which means `flash_attn` doesn't even have to be installed. 
Having AITER installed, xDiT correctly logs that AITER is installed and will be used as the attention library, but missing `flash_attn` then triggers another warning log:
```
Flash Attention library "flash_attn" not found, using pytorch attention implementation
```
SDPA will not be used if AITER is present, so this PR gates this warning, so only if AITER is not installed will this be logged.

### DIFF
--- a/xfuser/envs.py
+++ b/xfuser/envs.py
@@ -164,12 +164,12 @@ class PackagesEnvChecker:
         return cls._instance
 
     def initialize(self):
-        self.packages_info = {
-            "has_aiter": self.check_aiter(),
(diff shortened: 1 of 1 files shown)


feifeibear (approved):
LGTM

Issue #576, opened 2025-10-16
is xDiT support run wan2.1 and wan2.2 ?


ying2025:
Does xDiT support running wan2.1 and wan2.2? Is there any detailed documentation, including for ComfyUI operation.
@Steaunk @lpetre

feifeibear:
Wan2.1 already integrates xDiT natively, so there’s no need for us to add separate support.
Wan2.2 implements its own Ulysses-style parallelism, independently of our framework.

avjves:
We are currently porting Wan2.X from diffusers to xDiT for easier interoperabilty. I'll link the branch/PR when it's ready. Does this sound something that could be merged into upstream xDiT @feifeibear?

By the way, Wan2.1 repo no longer works OOB with latest xDiT due to PR #559 .

feifeibear:
> We are currently porting Wan2.X from diffusers to xDiT for easier interoperabilty. I'll link the branch/PR when it's ready. Does this sound something that could be merged into upstream xDiT [@feifeibear](https://github.com/feifeibear)?
> 
> By the way, Wan2.1 repo no longer works OOB with latest xDiT due to PR [#559](https://github.com/xdit-project/xDiT/pull/559) .


Wan2.2 support of course can be upstream to xDiT!

Wan2.1 repo no longer works OOB with latest xDiT. Because of the torch version?

avjves:
Great!

The PR #559 changed the default `attn_type` to None from `AttnType.FA`. None is not supported, so one _has_ to specify an attn_type to use it.

Wan2.1 relies on the old default behaviour here and doesn't specify the type:
https://github.com/Wan-Video/Wan2.1/blob/7c81b2f27defa56c7e627a4b6717c8f2292eee58/wan/distributed/xdit_context_parallel.py#L213

So running it errors here:
https://github.com/feifeibear/long-context-attention/blob/7a52abd669efb35e550680a239e1745b620b2bae/yunchang/kernels/__init__.py#L257

with error:
```
Unknown flash attention implementation: None
```

feifeibear:
@avjves thanks for pointing out the problem. Coujld you plese help to fix the error by reverting part of the code?

avjves:
This will then break the support for Apple Silicon that was introduced in that PR. I'm not familiar with it nor do I have access to Apple Silicon, so I can't fix that if I add a PR to revert that change. Is this ok?

feifeibear:
Of course, at least we can revert the PR.
[...]

PR #578, opened 2025-10-21
Add workaround for default attn_type

avjves:
## What?
Adds a workaround for allowing the default `attn_type` to be `AttnType.FA` for legacy reasons while still being able to run xDiT  without `yunchang` installed.

## Why?
As discussed in the issue #576 ,  for example Wan2.1 repo no longer works out-of-the-box with the xDiT, as it assumes the default value for `attn_type` to be `AttnType.FA`, which has been set to `None` by PR #559 . 

## How?
We set the default `attn_type` within init call itself, not as the default value. This keeps the support for Apple Silicon without interfering with legacy default values.

@feifeibear


feifeibear (approved):
LGTM!

PR #607, opened 2026-01-08
Deprecate Yunchang attention paths in favor of USP

avjves:
# What?
Deprecates the hybrid sequence parallel attn path that uses Yunchang and changes the calls to use USP.

# Why?
The Yunchang path did not support torch.compile, so it had to have torch.compile disabled around the attention call. This added an unnecessary graph break there, which caused a performance hit. The USP path doesn't have the same issue, so this overall gives a boost in perf to all affected models. This change makes all the models use the same attention dispatch mechanism, so the changes from PR #604 now affect all models.

# How?
Swapped all `hybrid_seq_parallel_attn()` calls to `USP()` in all attention processors. I didn't try to combine the logic trees in the attention processors and left the old logic there intact, so there's plenty of overlap that could be streamlined still in the future. Only logic change was that I collapsed the if/else logic that chose between FA and SDPA, as `USP` already does this choice.

### DIFF
--- a/xfuser/model_executor/layers/attention_processor.py
+++ b/xfuser/model_executor/layers/attention_processor.py
@@ -1218,7 +1218,11 @@ class xFuserSanaAttnProcessor2_0(SanaAttnProcessor2_0):
                 "Currently SANA not support split_text_embed_in_sp!"
             )
         else:
+            query = query.transpose(1, 2)
+            key = key.transpose(1, 2)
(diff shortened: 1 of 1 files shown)


avjves:
Re-added the fast attn check, though it's not in use anymore. Shapes are correct and the attention mask is None on all calls. Most methods don't support custom attention mask, so USP doesn't support it either.

PR #610, opened 2026-01-12
Fix Wan model after yunchang deprecation

avjves:
# What?
A small addition to #607 , removes the import/call to `set_hybrid_seq_parallel_attn`, which was removed. I forgot to remove these calls alongside when submitting the previous PR.

# Why?
This currently throws an error, as it cannot import this. The import/call in this function were already unnecessary, as the transformer calls USP, not hybrid_seq_parallel_attn, so these can be removed without worry.


jcaraban (approved):
CI/CD where are you...

PR #613, opened 2026-01-13
Ascend NPU supports Ulysses CP and Ring CP.

Feng0w0:
Add a `ring_npu_flash_attn.py` to implement NPU's ring cp logic under `xfuser/core/long_ctx_attention/ring`.
Add a branch in the initialization of `xFuserLongContextAttention` to support NPU as an optional backend.

### DIFF
--- a/xfuser/core/long_ctx_attention/hybrid/attn_layer.py
+++ b/xfuser/core/long_ctx_attention/hybrid/attn_layer.py
@@ -87,8 +87,12 @@ class xFuserLongContextAttention(LongContextAttention):
                 raise RuntimeError("Sparse Sage attention does not support ring degree > 1.")
 
         self.attn_processor = attn_processor
-        from xfuser.core.long_ctx_attention.ring import xdit_ring_flash_attn_func
-        self.ring_attn_fn = xdit_ring_flash_attn_func
+        if attn_type == AttnType.NPU:
+            from xfuser.core.long_ctx_attention.ring import xdit_ring_npu_flash_attn_func
+            self.ring_attn_fn = xdit_ring_npu_flash_attn_func
+        else:
+            from xfuser.core.long_ctx_attention.ring import xdit_ring_flash_attn_func
+            self.ring_attn_fn = xdit_ring_flash_attn_func
 
     @torch.compiler.disable
     def forward(
--- a/xfuser/core/long_ctx_attention/ring/__init__.py
+++ b/xfuser/core/long_ctx_attention/ring/__init__.py
@@ -2,8 +2,10 @@ from .ring_flash_attn import (
     xdit_ring_flash_attn_func, 
     xdit_sana_ring_flash_attn_func,
 )
+from .ring_npu_flash_attn import xdit_ring_npu_flash_attn_func
 
 __all__ = [
     "xdit_ring_flash_attn_func",
     "xdit_sana_ring_flash_attn_func",
+    "xdit_ring_npu_flash_attn_func",
 ]
--- a/xfuser/core/long_ctx_attention/ring/ring_npu_flash_attn.py
+++ b/xfuser/core/long_ctx_attention/ring/ring_npu_flash_attn.py
@@ -1,0 +1,247 @@
+from typing import List
+import math
+import torch
+import torch.nn.functional as F
+
+from xfuser.core.long_ctx_attention import xFuserLongContextAttention
+from xfuser.core.cache_manager.cache_manager import get_cache_manager
+import xfuser.envs as envs
+
+if torch.cuda.is_available() or envs._is_npu():
+    from yunchang.ring.utils import RingComm, update_out_and_lse, update_npu_out
+    from yunchang.ring.ring_npu_flash_attn import RingNpuFlashAttnFunc
+    from yunchang.kernels import select_flash_attn_impl, AttnType
+else:
+    RingComm = object
+    RingNPUFlashAttnFunc = object
+    AttnType = None
+    select_flash_attn_impl = None
+
+try:
+    import flash_attn
+    from flash_attn.flash_attn_interface import _flash_attn_forward
+except ImportError:
+    flash_attn = None
+    _flash_attn_forward = None
+
+
+def xdit_ring_npu_flash_attn_forward(
+        process_group,
+        q: torch.Tensor,
+        k: torch.Tensor,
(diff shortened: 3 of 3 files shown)


avjves:
Hi! This PR adds changes to the yunchang/long context attention path. This has been deprecated in favor of using the `USP` method due to performance reasons and simplicity. Would it be possible for you to use [USP ](https://github.com/xdit-project/xDiT/blob/77e8b781c590f062bda65a9fa6f28688cca3142c/xfuser/model_executor/layers/usp.py#L225)instead? The backends are now defined in the `attention_backend.py` file [here](https://github.com/xdit-project/xDiT/blob/main/xfuser/core/distributed/attention_backend.py).

I believe the NPU attention backend code is this:
```
def npu_fused_attn_forward(q, k, v, 
        head_num = None, 
        input_layout = "BSND",  
        scale = None, 
        pre_tokens=65535, 
        next_tokens=65535
        ):
    assert HAS_NPU, "torch_npu is not avaliable"
    if scale is None:
        scale = q.shape[-1] ** -0.5
    attention_out, softmax_max, softmax_sum,_,_,_,_ = torch_npu.npu_fusion_attention_v2(q, k, v, 
                                                head_num = head_num, 
                                                input_layout = input_layout,  
                                                scale = scale, 
                                                pre_tokens=pre_tokens, 
                                                next_tokens=next_tokens)
    lse = torch.logsumexp(attention_out, dim=-1)
    # print(f"lse shape is: {lse.shape}, softmax_sum shape is: {softmax_sum.shape}, softmax shape is: {softmax_max.shape}")
    return attention_out, softmax_max, softmax_sum, scale
```

This can be copied as an attention backend similarily to the others there almost verbatim. The QKV values provided by USP are in a different shape, so they need to be transposed. So the implementation would be:

```
@register_attention_function(AttentionBackendType.NPU)
def npu_fused_attn_forward(q, k, v, 
        head_num = None, 
        input_layout = "BSND",  
        scale = None, 
        pre_tokens=65535, 
        next_tokens=65535
        ):
    assert HAS_NPU, "torch_npu is not avaliable"
    q = q.transpose(1, 2)
    k = k.transpose(1, 2)
    v = v.transpose(1, 2)
    if scale is None:
        scale = q.shape[-1] ** -0.5
    attention_out, softmax_max, softmax_sum,_,_,_,_ = torch_npu.npu_fusion_attention_v2(q, k, v, 
                                                head_num = head_num, 
                                                input_layout = input_layout,  
                                                scale = scale, 
                                                pre_tokens=pre_tokens, 
                                                next_tokens=next_tokens)
    lse = torch.logsumexp(attention_out, dim=-1)
    # print(f"lse shape is: {lse.shape}, softmax_sum shape is: {softmax_sum.shape}, softmax shape is: {softmax_max.shape}")
    return attention_out, softmax_max, softmax_sum, scale
```

and then add that AttentionBackendType.NPU to the Enum in the same file.


For NPU you have guarded some imports behind cuda availability, so I'm not sure if USP.py file requires some of those as well.

avjves:
Does NPU support this?
```
from torch.distributed.tensor.experimental._attention import _templated_ring_attention
```
`USP` uses that for Ring. If not, then we have to use `LongContextAttention` with the above fixes for NPU devices and the rest will use `USP`.

Feng0w0:
@avjves Thank you very much for your reply, but currently NPU does not support using the `_templated_ring_attention` function because the `npu_fused_attn_forward` function used by NPU returns `out, softmax_max, softmax_sum`. instead of `out, softmax_lse`.And I understand that `_templated_ring_attention` function needs to ensure that the passed `attention_function` returns `out, softmax_les`

avjves:
I see, that's unfortunate. And it can't be calculated from the output values? I can see in yunchang you used to have lse there as well, but was that not working?

If not possible, then we have to still support the two paths, USP and Yunchang. We should add a check in the `envs.py` to determine whether to use USP for attention or Yunchang. Basically for now this would check if one uses NPU or not. Then all attention processors should be updated to call a new helper function `dispatch_attention` (for example) that would then check the value from envs and forward the call to the correct attention path.

@Feng0w0 if you are not in a super hurry, I can do the attention related changes in a different PR and when that is merged we can make sure this PR works well with the changes. Or if you are in a hurry, you are welcome to do the changes yourself as well :)

This PR also cannot be merged before https://github.com/feifeibear/long-context-attention/pull/169/files is merged, as the imports would break xDiT currently :)

Feng0w0:
@avjves If xDiT only supports inference, NPU has a function `torch_npu.npu_fused_infer_attention_score` that can return `out, lse`, so it can be integrated into the USP interface. However, I am currently unsure how to use multiple cards to test the USP interface. Do you have any use cases to refer to?

Feng0w0:
May I ask if it is currently impossible to use the USP interface separately from the xDit library for Attn calculation? Before using USP, it is necessary to call initialize_runtime_state to initialize _RUNTIME, which must receive the pipeline. Previously, the xFuserLongContext interface could be used without initializing the _RUNTIME.

avjves:
xDiT is purely an inference engine, so no need to worry about training / backward pass.

USP has support for some stateful attention schemes, so a runtime is required. However, you don't have to provide full pipeline if you just want to call the USP methout outside of xDiT for testing:

```
from xfuser.core.distributed import (
    initialize_model_parallel
    initialize_runtime_state
)

from xfuser.model_executor.layers.usp import USP

initialize_runtime_state()
initialize_model_parallel(ring_degree=1, ulysses_degree=1)
USP(q, k, v, drouput, is_causal) # Attention call

```

But for multi-GPU testing I just recommend running some model with xDiT normally. All the comms etc. are then handled automatically and you only need to add the NPU backend.

Feng0w0:
@avjves Thank you very much for your reply. I found that there is a open-source repository using the xFuserLongContext interface to implement its USP feature. If possible, could you please merge this PR first?  This PR can be used to call the Attn calculation of NPU through the xFuserLongContext interface. And I will submit another PR later to add npu_flash_attn as an optional backend for the USP interface. In this way, both the xFuserLongContext interface and USP interface can choose NPU for Attn calculation.

avjves:
Sounds good! I naturally haven't tested these changes myself as I don't have access to NPUs, but they seem good :) This is nicely also separated from non-NPU cases.

avjves (approved):

PR #620, opened 2026-01-20
Remove graph break before aiter fp8 call

ksikiric:
Due to a bug with the fake tensors in AITER, the fp8 calls crashed due to dtype mismatch when using torch compile. A workaround was to explicitly cause a graph break before the attention call to force dtype materialization. This issue has now been resolved in AITER, and we can now remove the graph break, increasing fp8 perf with 2%.

Also added a sentence in the README, to warn users with older AITER versions about the potential issue when using fp8 attention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ Several different attention backends are supported:
 
 xDiT comes with `flash_attn` as an optional install requirement, as it currently supports the largest variety of different GPU architectures.
 However, newer implementations generally offer better performance. If available for you, we highly recommend using `cuDNN`, `FAv3` (on _hopper_ GPUs) or `FAv4` (on _blackwell_ GPUs).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,7 +250,7 @@ Several different attention backends are supported:
 
 xDiT comes with `flash_attn` as an optional install requirement, as it currently supports the largest variety of different GPU architectures.
 However, newer implementations generally offer better performance. If available for you, we highly recommend using `cuDNN`, `FAv3` (on _hopper_ GPUs) or `FAv4` (on _blackwell_ GPUs).
-On recent AMD GPUs (MI300X or newer) it is generally recommended to use `AITER` in all cases to get the best possible performance.
+On recent AMD GPUs (MI300X or newer) it is generally recommended to use `AITER` in all cases to get the best possible performance. Note that when using `AITER FP8` as the attention backend with `torch.compile`, it is important to use the latest version of `AITER`, as older versions may trigger a fake tensor-related bug, resulting in a runtime error.
```
jcaraban: can you specify a version number and/or date (e.g. January 2026) on the AITER note?

PR #624, opened 2026-01-26
Hybritd attn to unified runner

ksikiric:
Hybrid attention was omitted from the first unified runner PR, adding it here. Currently only enabled for HunyuanVideo and Wan.

See: https://github.com/xdit-project/xDiT/pull/608 for motivation behind Hybrid Attention

### DIFF
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -622,7 +622,7 @@ class xFuserArgs:
             "--num_hybrid_bf16_attn_steps",
             type=int,
             default=None,
-            help="Number of steps to use BF16 attention."
+            help="Number of BF16 steps to use with hybrid attention attention."
         )
         return parser
 


xfuser/config/args.py
```diff
@@ -610,6 +612,18 @@ def add_runner_args(parser: FlexibleArgumentParser):
...
+            "--use_hybrid_fp8_attn",
+            action="store_true",
+            default=False,
+            help="Enable hybrid FP8 attention for faster inference and improved quality."
+        )
+        parser.add_argument(
+            "--num_hybrid_bf16_attn_steps",
+            type=int,
+            default=None,
+            help="Number of steps to use BF16 attention."
```
avjves: Nit: Specify here that `num_hybrid_bf16_attn_steps` is only used when `use_hybrid_fp8_attn` is enabled. As models just ignore it silently otherwise, it's better for the help text to at least specify why it does that.

PR #628, opened 2026-01-27
Add optimized group norm from AITER when available

ksikiric:
AITER supports an optimized group norm kernel that improves Hunyuan perf by 5%. This PR adds this optimized group norm as default whenever available, otherwise fall back to torch.nn.GroupNorm. It is also used by Flux and Z-image, although performance gains are not as prominent as with Hunyuan. 

Changes
- Add conditional import and usage of AITER's optimized GroupNorm
- Maintain backward compatibility with automatic fallback to PyTorch's native implementation
- Default compilation behavior of the unified runner should be to compile only the transformers to match the legacy examples

### DIFF
--- a/xfuser/envs.py
+++ b/xfuser/envs.py
@@ -200,15 +200,6 @@ class PackagesEnvChecker:
         """
         try:
             import aiter
-            try:
-                from aiter.ops.groupnorm import GroupNorm
-                torch.nn.GroupNorm = GroupNorm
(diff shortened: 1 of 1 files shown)


avjves:
I think for clarity the groupnorm setting could be moved from the package checker into it's own function after it, e.g. `_setup_rocm_libraries()` which then checks the presense of aiter from the package checker prior to adding the group norm :)

PR #641, opened 2026-02-06
Remove deprecated examples

avjves:
# What?
Removes example scripts from models that can be ran with the xdit model runner.

# Why?
Having two different ways of running the same model is problematic, especially if they don't fully share the code:
1. Adding new features requires adding them to two places, creating unncessary work
2. The example scripts are not actively tested anymore, so at any point they could become unfunctional
3. Users might use deprecated scripts that do not work -> bad PR


jcaraban:
shall this PR extend the README so it is clear how to use the `xdit` model runner?
_edit:_ right, I had not seen PR https://github.com/xdit-project/xDiT/pull/642 with does exactly that, all good

PR #664, opened 2026-03-06
Mxfp4 gemm scheduler

ksikiric:
Wan2.2-5B (dense) is more sensitive to quality degradation with MXFP4 GEMMs compared to Wan2.2-14B (MoE), and it's challenging to maintain good video quality even by excluding early or last layers from mxfp4 quantization. However, diffusion-step scheduling similar to hybrid-attention seems to work with GEMM quantization.

Args:

- `--use_hybrid_gemm_schedule turns` on step-wise GEMM precision switching
- `--num_hybrid_gemm_high_precision_steps N`: sets how many higher-precision FP8 steps are used at both ends of denoising
- Requires  `--use_fp4_gemms flags`
- Behavior is FP8 (first N) → MXFP4 (middle) → FP8 (last N)

This way, we can achieve a 4% perf improvement over FP8 Gemms (single GPU run) and 2.5% on 8 GPU run.

Output from single GPU run with below command:
`torchrun --nproc_per_node=1 xfuser/runner.py --height 736 --width 1280 --num_frames 121 --model Wan2.2-TI2V --ulysses_degree 1 --prompt "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds. The cat assumes a naturally relaxed posture, as if savoring the sea breeze and warm sunlight. A close-up shot highlights the feline's intricate details and the refreshing atmosphere of the seaside." --input_images https://huggingface.co/datasets/YiYiXu/testing-images/resolve/main/wan_i2v_input.JPG --use_torch_compile --num_inference_steps 50 --use_hybrid_gemm_schedule --use_fp4_gemms --task i2v --resize_input_images `

https://github.com/user-attachments/assets/c823ead1-fc78-4ed4-bdc1-ba9bff9d1584


xfuser/core/distributed/attention_schedule.py
```diff
@@ -74,4 +74,42 @@ def create_hybrid_attn_schedule(
...
+
+
+class GemmPrecisionSchedule:
+    """
+    Per-step GEMM precision schedule defined by an explicit list of booleans.
+    use_high_precision[i] indicates whether step i should use high precision GEMM.
+    """
+
+    def __init__(self, use_high_precision: List[bool]):
+        if not use_high_precision:
```
avjves: nit: this sounds like a boolean value variable, not a list. Maybe high_precision_steps would be better?

xfuser/core/distributed/attention_schedule.py
```diff
@@ -74,4 +74,42 @@ def create_hybrid_attn_schedule(
...
+            raise ValueError("GemmPrecisionSchedule requires at least one step.")
+        self.use_high_precision = list(use_high_precision)
+        self.total_steps = len(self.use_high_precision)
```
avjves: Maybe I'm missing something, but is this used anywhere? And why is total steps the same as high precision steps? 🤔

xfuser/core/distributed/attention_schedule.py
```diff
@@ -74,4 +74,42 @@ def create_hybrid_attn_schedule(
...
+    """
+    Create a hybrid GEMM schedule: high-precision GEMM at start/end, low-precision GEMM in the middle.
+    """
+    num_low_precision_steps = total_steps - 2 * num_high_precision_steps
+    if num_low_precision_steps < 0:
+        raise ValueError(
+            f"total_steps ({total_steps}) must be >= 2 * num_high_precision_steps ({2 * num_high_precision_steps})."
+        )
+    schedule = (
+        [True] * num_high_precision_steps
```
avjves, line 110: Would we benefit from having this be similar to attention schedule, where it's not just a true/false for FP8/MXFP4?

So one could similarly specify the scheduler, i.e:
```
--hybrid_gemm_schedule BF16,FP8,MXFP4,FP8,BF16
```
?

Not sure how similar this is to attn backends where we might end up with different ones in the future. If this doesn't sound relevant, ignore

xfuser/core/distributed/runtime_state.py
```diff
@@ -272,18 +275,43 @@ def has_attention_schedule(self) -> bool:
...
+            if self.schedule_total_steps != self.gemm_schedule_total_steps:
+                raise RuntimeError(
+                    f"Attention and GEMM schedules must use the same total steps; got {self.schedule_total_steps} and {self.gemm_schedule_total_steps}."
+                )
+            return self.schedule_total_steps
+        if self.schedule_total_steps is not None:
+            return self.schedule_total_steps
+        return self.gemm_schedule_total_steps
+
+    @torch._dynamo.disable
```
avjves: What does this do? Does it cause a graph break here? This function is called even without having enabled the hybrid schedules, so would cause a regression for standard use 🤔

xfuser/core/distributed/runtime_state.py
```diff
@@ -297,10 +325,24 @@ def set_attention_schedule(
         for backend in set(attention_schedule.backends):
             self._check_if_backend_compatible_with_current_configuration(backend)
         self.attention_schedule = attention_schedule
-        self.schedule_total_steps = torch.tensor(total_steps, dtype=torch.int)
-        self.step_counter = torch.tensor(0, dtype=torch.int)
+        self.schedule_total_steps = total_steps
```
avjves: This no longer causes a graph break?

xfuser/core/distributed/runtime_state.py
ksikiric: hm, I missed this... I'll remove it

xfuser/core/distributed/runtime_state.py
ksikiric: Probably due to the above _dynamo.disable, will revert this

xfuser/core/distributed/attention_schedule.py
ksikiric: I think this is only needed for fp4 gemms, as fp8 gemms already produce good quality. And combining fp4 with bf16 would not gain significantly (if at all) to using pure fp8 gemms.

avjves (approved):

PR #694, opened 2026-04-24
Enable LTX-2.3

kTorp:
# What

Add LTX-2.3 support to xDiT, following the new 2.3-specific features introduced in upstream diffusers PR
[huggingface/diffusers#13217](https://github.com/huggingface/diffusers/pull/13217) 

# Why

xDiT already supports LTX-2.0 but has no runner for LTX-2.3.

# How

- Adds `xFuserLTX23VideoModel`, a runner class driving a two-stage up-scaling pipeline.
- Extends `xFuserLTX2VideoTransformer3DWrapper` to mirror the upstream transformer
- Ports `xFuserLTX2PerturbedAttnProcessor` from upstream diffusers.

# Limitations

STG and modality isolation each add an extra full transformer forward per step (inherent to the upstream pipeline design), which adds overhead. These extra steps are opt-in via pipeline flags and can be enabled/disabled independently.

Need updated diffusers version.

# Tests

On MI300
https://github.com/user-attachments/assets/2f176fe7-a53e-4cb6-9b97-e887db70f827

On B300
https://github.com/user-attachments/assets/50ba66b1-8d23-439c-aeeb-49a2159fd442

LTX2.0 on MI300

https://github.com/user-attachments/assets/15bb0700-9d2d-4fe2-b539-5a363d59478f


xfuser/config/args.py
```diff
@@ -547,6 +547,55 @@ def add_runner_args(parser: FlexibleArgumentParser):
             type=float,
             help="Flow shift for the scheduler.",
         )
+        parser.add_argument(
```
avjves: Do we need to have these as configurable? 🤔 They seem highly specific to LTX2.3 and pollute the CLI args quite a lot. In CausalWan I did these like this:
https://github.com/xdit-project/xDiT/blob/051db68f769efa5b1aa691e5eec133f1b8bf3f1f/xfuser/model_executor/models/runner_models/causal_wan.py#L48

xfuser/model_executor/models/runner_models/ltx.py
```diff
@@ -33,6 +33,162 @@
...
+            audio_latents=audio_latent,
+            prompt=input_args["prompt"],
+            negative_prompt=input_args["negative_prompt"],
+            height=input_args["height"],
+            width=input_args["width"],
+            num_frames=input_args["num_frames"],
+            frame_rate=self.settings.fps,
+            num_inference_steps=3,
+            noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0],
+            sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
```
avjves, line 157: Just checking, these are shared between LTX2 and 2.3?

xfuser/config/args.py
kTorp: Likely not, and they do end up polluting the CLI with LTX2.3 specific arguments, so the CausalWan method looks like a good way to handle this. I'll make the changes 😄

xfuser/model_executor/models/runner_models/ltx.py
kTorp: I believe so, for the 2-stage pipeline I looked at the 2-stage (I2V) example in the diffusers PR: https://github.com/huggingface/diffusers/pull/13217 which imports `from diffusers.pipelines.ltx2.utils import STAGE_2_DISTILLED_SIGMA_VALUES`

avjves:
Can you provide videos from LTX2 as well? Just to make sure it still works as intended 😄

avjves (approved):

PR #696, opened 2026-05-04
Use ATen-level distributed optimizations for Inductor compute/comm overlap

lauri9:
### Problem

Every runner model toggles Inductor compute/communication overlap by hand
right before `torch.compile`:

```python
torch._inductor.config.reorder_for_compute_comm_overlap = True
self.pipe.transformer = torch.compile(...)
```

This has two issues:

1. **Duplication.** The same line is copy-pasted across `base_model.py`,
   `flux.py`, `hunyuan.py` (×2), `ltx.py` (×2), `wan.py` (×5) and
   `causal_wan.py`. Any change to the policy has to be made in nine places.
2. **Stale on PyTorch 2.10+.** PyTorch upstream now recommends the new
   ATen-level distributed optimization passes
   (`torch._inductor.config.aten_distributed_optimizations`) over the legacy
   scheduler-IR passes — see
   [pytorch/pytorch#169461](https://github.com/pytorch/pytorch/issues/169461).
   On top of that, PyTorch 2.10 emptied the default value of
   `reorder_for_compute_comm_overlap_passes` (it used to be
   `["reorder_compute_for_overlap", "sink_waits", "raise_comms"]` on 2.9), so
   simply flipping `reorder_for_compute_comm_overlap = True` on a 2.10 build
   is a no-op and silently regresses overlap.

### Fix

Add a `configure_inductor_comm_overlap()` helper in
`xfuser/core/utils/runner_utils.py` and call it from every runner model's
`_compile_model`.

The helper:

- **Prefers the ATen passes** when
  `torch._inductor.config.aten_distributed_optimizations` is available
  (torch ≥ 2.10). It enables `enable_overlap_scheduling`,
  `collective_bucketing` and `insert_overlap_deps`, and forces
  `collective_estimator="benchmark"`. The benchmarking estimator is required
  on platforms such as MI355X, where the analytical NCCL bandwidth model can be ~10×
  off and leaves the scheduler thinking collectives are far more expensive
  than they actually are, suppressing overlap.
- **Disables the legacy scheduler-IR overlap passes** in the same path, so
  the ATen reorderer and the scheduler reorderer don't fight over the same
  graph.
- **Falls back to the legacy passes** on older torch builds, and explicitly
  restores the pre-2.10 default
  `reorder_for_compute_comm_overlap_passes = ["reorder_compute_for_overlap",
  "sink_waits", "raise_comms"]` so 2.10 builds without the ATen API still
  get real overlap instead of an empty pass list.

### Changes

- `xfuser/core/utils/runner_utils.py` — new
  `configure_inductor_comm_overlap()` helper with ATen-pass preference and
  legacy fallback.
- `xfuser/model_executor/models/runner_models/base_model.py`
- `xfuser/model_executor/models/runner_models/causal_wan.py`
- `xfuser/model_executor/models/runner_models/flux.py`
- `xfuser/model_executor/models/runner_models/hunyuan.py` (HunyuanVideo and HunyuanVideo-1.5)
- `xfuser/model_executor/models/runner_models/ltx.py` (LTX-2 and LTX-2.3)
- `xfuser/model_executor/models/runner_models/wan.py` (Wan2.1/Wan2.2 I2V, T2V, TI2V)

  All `_compile_model` methods now call `configure_inductor_comm_overlap()`
  instead of setting `torch._inductor.config.reorder_for_compute_comm_overlap`
  directly.

### DIFF
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -18,6 +18,29 @@ def log(message: str, debug=False, log_from_all_processes: bool = False) -> None
         else:
             logger.info(message)
 
+def configure_inductor_comm_overlap() -> None:
+    """Enable Inductor compute/communication overlap.
+
+    PyTorch 2.10 emptied the default value of
+    ``torch._inductor.config.reorder_for_compute_comm_overlap_passes`` (it
+    used to be ``["reorder_compute_for_overlap", "sink_waits", "raise_comms"]``
+    on 2.9). With the list empty, even though
+    ``reorder_for_compute_comm_overlap`` is on, no scheduler-IR reorder
+    runs and bucketed collectives stay issued in source order, so the
+    runtime never gets a chance to overlap them with compute.
+
+    Restoring the 2.9 default list here is a no-op on torch <= 2.9
+    (the value is already that list) and recovers the 2.9 overlap
+    behavior on torch >= 2.10.
+    """
+    torch._inductor.config.reorder_for_compute_comm_overlap = True
+    torch._inductor.config.reorder_for_compute_comm_overlap_passes = [
+        "reorder_compute_for_overlap",
+        "sink_waits",
+        "raise_comms",
+    ]
+
+
 def is_last_process() -> bool:
     """
     Checks based on env rank and world size if this is last process in
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -21,18 +21,43 @@ def log(message: str, debug=False, log_from_all_processes: bool = False) -> None
 def configure_inductor_comm_overlap() -> None:
     """Enable Inductor compute/communication overlap.
 
-    PyTorch 2.10 emptied the default value of
-    ``torch._inductor.config.reorder_for_compute_comm_overlap_passes`` (it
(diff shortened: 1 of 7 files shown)


jcaraban:
> ... and forces collective_estimator="benchmark"

did you check how much extra time the benchmark adds, or is it negligible?

avjves:
So this now works with the old and new torch versions?

jcaraban (approved):

lauri9:
> > ... and forces collective_estimator="benchmark"
> 
> did you check how much extra time the benchmark adds, or is it negligible?

It is negligible in comparison. On PyTorch 2.9, with compile overhead my Wan 2.2 workload finished in 319s. On PyTorch 2.11 with this PR it was 316s.

lauri9:
> So this now works with the old and new torch versions?

Yes, it still works on 2.9 which I tested just now. You then get the printout:
```
DEBUG 05-04 11:37:36 [runner_utils.py:17] torch._inductor.config.aten_distributed_optimizations is unavailable on this torch build; using legacy scheduler-IR overlap passes.
```
As the new ATen FX pass is not available on this build.

I compared main and this PR on torch 2.9, and runtime was within 0.1s on a Wan 2.2 workload.

PR #699, opened 2026-05-08
feat: FSDP2 sharding with quantized GEMMs support

pds-amd:
Adds FSDP2 (`fully_shard`) support to xDiT's sharding pipeline, enabling `torchao` quantized GEMMs to work correctly under distributed sharding. Previously, quantization was applied after FSDP1 wrapping, which is incompatible with `torchao` tensor subclasses that cannot be flattened to FSDP1's 1D `FlatParameter`.

  - `runner_utils.py`: Patches `torchao`'s `Float8Tensor` for FSDP2 compatibility. Fixes `aten.split`, `new_empty`, `new_zeros`, `copy_`, view, `as_strided`, and adds `fsdp_pre`/`post_all_gather` so FSDP2 all-gather correctly reconstructs `Float8Tensor` subclasses instead of returning raw fp8 bytes.                                
  - `sharding.py`: `shard_component` now accepts `quantize_fn` and `reshard_after_forward` params. When `quantize_fn` is provided, takes the FSDP2 path: quantizes each block in-place before `fully_shard`, loading one block at a time to minimise peak memory. `fully_shard` is imported lazily to avoid registering composable FSDP dispatch machinery for non-quantized workloads.                                                                                                                                                                                                                                                                 
  - `base_model.py`: `_shard_model_with_fsdp` wires up per-block quantization via `_build_fsdp_quantize_fn`. Non-FSDP quantization paths (`_setup_nvfp4_gemms`, `_setup_mxfp4_gemms`) are unchanged.
  - `args.py`: Adds `--no_reshard_after_forward` flag.
  
  Tested extensively with Flux2 and Wan2.2.

### DIFF
--- a/xfuser/core/distributed/sharding.py
+++ b/xfuser/core/distributed/sharding.py
@@ -13,10 +13,11 @@ Key Features:
 Functions:
     - shard_dit: Shard a Diffusion Transformer (DiT) model
     - shard_t5_encoder: Shard a T5 encoder model
-    - shard_transformer_blocks: Generic transformer block sharding
+    - shard_component: Generic transformer block sharding
 """
+import logging
 from functools import partial
-from typing import Iterable, Optional, Any
+from typing import Callable, Iterable, Optional
 
 import torch
 import functools
@@ -25,9 +26,21 @@ from torch.distributed.fsdp import (
     ShardingStrategy,
     FullyShardedDataParallel as FSDP,
 )
-from torch.distributed.fsdp.wrap import (
-    lambda_auto_wrap_policy,
-)
+from torch.distributed.fsdp.wrap import lambda_auto_wrap_policy
+from torch.distributed.device_mesh import DeviceMesh
+
+
+logger = logging.getLogger(__name__)
+
+
+def _make_mesh(
+    process_group: Optional[torch.distributed.ProcessGroup],
+    device_type: str = "cuda",
+):
+    """Wrap an existing ProcessGroup as a 1-D DeviceMesh without creating a new NCCL communicator."""
+    if process_group is None:
+        return None
+    return DeviceMesh.from_group(process_group, device_type)
 
 
 def children_to_device(
@@ -55,7 +68,7 @@ def children_to_device(
     Example:
         >>> model = TransformerModel()
         >>> # Move all children except 'blocks' to GPU
-        >>> _children_to_device(model, 'cuda:0', excluded_children=['blocks'])
+        >>> children_to_device(model, 'cuda:0', excluded_children=['blocks'])
     """
     for name, child in module.named_children():
         if name not in excluded_children:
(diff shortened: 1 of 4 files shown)


xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -535,49 +545,100 @@ def _shard_model_with_fsdp(self) -> None:
...
+        )
+
+        if not use_fp4_here and not use_fp8_here:
+            return None
+
+        def quantize_fn(block, block_idx: int) -> None:
+            block_prefix = f"{block_idx}."
+            # Strip the block-index prefix so the quantize functions see local FQN paths.
+            local_fp8 = tuple(
+                o[len(block_prefix):] for o in fp8_overrides if o.startswith(block_prefix)
```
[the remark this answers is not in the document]
pds-amd: block_prefix is e.g. "1." so the trailing dot prevents false matches on 10+ blocks:                                                                       

```
"1.attn".startswith("1.")  = True
"11.attn".startswith("1.")  = False
```

xfuser/core/utils/runner_utils.py
```diff
@@ -135,6 +135,171 @@ def _get_fp8_kernel_preference():
...
+    return patched
+
+
+_TORCHAO_FLOAT8_FSDP2_PATCHES: list[str] = []
+
+try:
+    _TORCHAO_FLOAT8_FSDP2_PATCHES = _patch_torchao_float8_fsdp2()
+    logger.debug("torchao Float8Tensor FSDP2 patches applied: %s", _TORCHAO_FLOAT8_FSDP2_PATCHES)
+except Exception as e:
+    logger.debug("torchao Float8Tensor FSDP2 patches skipped (%s): %s", type(e).__name__, e)
```
avjves, line 255: If this fails, does the whole quantization just crash? I'm just wondering if this should be a warning instead, as we are not running with debug enabled, so this would silently regress 🤔

avjves: Ah, the assertion later does it - ignore :)

avjves (approved):
Looks good, can't find anything wrong with it, though it's quite complex 😄 You verified it now allows the FP8 with sharding - did you verify the output as well?

xfuser/core/distributed/sharding.py
```diff
@@ -15,19 +15,26 @@
...
-    lambda_auto_wrap_policy,
-)
+from torch.distributed._composable.fsdp import fully_shard
+from torch.distributed.device_mesh import DeviceMesh
+from xfuser.core.utils.runner_utils import rgetattr
+
+logger = logging.getLogger(__name__)
+
+
+def _make_mesh(
```
jcaraban, line 36: I think this function might be throwing the following warning now
> /usr/local/lib/python3.12/dist-packages/torch/distributed/distributed_c10d.py:2776: UserWarning: `_get_pg_default_device` will be deprecated, it only stays for backward-compatibility reason. If you need to find a device for object collectives, please use `_get_object_coll_device`. If you need to query the device types supported by group, please use `_device_capability(group)`.
  device = device or _get_pg_default_device(group)

pds-amd: This is torch internal.

From what I can tell:

  1. FSDP2 introduces all-gather/reduce-scatter collective ops into the model graph                                                                                                                                                                                                                                           
  2. torch.compile runs Inductor on that graph
  3. configure_inductor_comm_overlap() enables Inductor's comm analysis on those collectives                                                                                                                                                                                                                                                                  
  4. _nccl_estimate() fires on each FSDP2 collective which calls _get_pg_default_device(pg) that produces the warning     

https://github.com/pytorch/pytorch/pull/175896 removes the `_get_pg_default_device` call that triggers the warning so a later version of pytorch won't emit this. Should we suppress the warning in xDiT or do we leave as is?

pds-amd:
> Looks good, can't find anything wrong with it, though it's quite complex 😄 You verified it now allows the FP8 with sharding - did you verify the output as well?

Yes, checked videos for each. 👍

xfuser/core/distributed/sharding.py
jcaraban: no need to suppress

pds-amd:
Switching to draft until I have performed more extensive profiling across different model families and configurations to assess how the fsdp change impacts latency, etc.

jcaraban (approved):

PR #702, opened 2026-05-14
Fix torch.cuda.get_device_properties() call for PyTorch < 2.6.0 compatibility

cszhouY:
## what
`torch.cuda.get_device_properties()` without arguments was introduced in PyTorch 2.6.0. Calling it without the device argument on earlier versions raises an error. This patch explicitly passes `torch.cuda.current_device()` as the device argument, which is supported across all PyTorch versions.


avjves (approved):
LGTM! Thanks :)

PR #707, opened 2026-05-21
feat: FSDP support for remaining models, memory-efficient init, and CPU offload fix

pds-amd:
## Summary

Four related changes that together unlock multi-GPU and memory-constrained inference for a broader set of models.

### 1. FSDP for most runner models

Adds `fully_shard_degree` capability and `fsdp_strategy` to 13 model classes that previously lacked FSDP support: FLUX.1-dev, FLUX.1-Kontext, FLUX.2-klein-9B/4B, Wan2.1-VACE, CausalWan, StableDiffusion-3.5, Z-Image/Turbo.

HunyuanVideo (all variants) & LTX-2/2.3 will be addressed in a future PR once some additional testing can be done.

The change per model is purely declarative. `fully_shard_degree=True` in `ModelCapabilities` and an `fsdp_strategy` dict that names the block attributes to wrap. The sharding logic itself lives in `base_model.py` and is model-agnostic.

Without FSDP, running large models across multiple GPUs is not feasable with lower VRAM GPUs.

### 2. Automatic text encoder sharding when `fully_shard_degree > 1`

Large text encoders (T5-XXL, Mistral3-7B, Gemma3, Qwen2.5-VL, LlamaModel) now appear in each model's `fsdp_strategy` and get FSDP-wrapped alongside the transformer. This reduces per-rank memory proportionally. A 7B text encoder across 4 GPUs costs ~1.75 GB/rank instead of ~7 GB/rank.

### 3. `--memory_efficient_sharding`

Routes FSDP init through FSDP2's sequential block-by-block loading, keeping at most one transformer block on GPU at a time during initialization. FSDP1 remains the default unless quantizing to maintain prior behavior unless this is specifically set.

### 4. Fix: skip `pipe.to(cuda)` when CPU offload is configured

`--enable_model_cpu_offload` and `--enable_sequential_cpu_offload` manage their own device placement via hooks. This allows larger models that would not fit on 1 GPU without offloading run now.

### 5. Text encoder FSDP wrapping

When using FSDP text encoders will also be sharded now. Note: several text encoders break when FSDP-wrapped per-layer due to `isinstance` checks in their forward pass. These are now wrapped as single FSDP units (`wrap_attrs: []`).

### 6. Mistral3 vision_tower kept on CPU for FLUX.2

FLUX.2's Mistral3 text encoder contains a SigLIP vision tower (~4 GB BF16) that is never invoked during text-only embedding. It is now moved to CPU before FSDP sharding, recovering ~4 GB per rank.

### 7. VRAM logging + `empty_cache` between FSDP component sharding steps

Logs `torch.cuda.memory_allocated()` after each component shard and calls `torch.cuda.empty_cache()` to release freed allocator blocks before the next component, reducing peak VRAM during FSDP init.

### 8. CPU offload for text encoders via FSDP2 `CPUOffloadPolicy`

Adds `offload_policy: cpu` support in `fsdp_strategy`. When set, text encoder shards are kept on CPU between uses and only all-gathered to GPU during `encode_prompt`. This eliminates the text encoder all-gather buffer (~14 GB for Mistral3) competing with the transformer during multi-GPU inference.

### DIFF
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -604,7 +604,8 @@ class xFuserModel(abc.ABC):
         device_group = get_fs_group().device_group
         for component_name, component in self.pipe.components.items():
             if component_name in self.settings.fsdp_strategy:
-                log(f"Sharding {component_name} with FSDP...")
+                log(f"Sharding {component_name} with FSDP... "
+                    f"(VRAM before: {torch.cuda.memory_allocated(local_rank)/1e9:.2f}GB)")
                 strategy = self.settings.fsdp_strategy[component_name]
                 wrap_attrs = strategy.get("wrap_attrs", [])
                 dtype = strategy.get("dtype", None)
@@ -617,8 +618,11 @@ class xFuserModel(abc.ABC):
                     memory_efficient_init=self.config.memory_efficient_sharding,
                 )
                 setattr(self.pipe, component_name, fsdp_object)
+                torch.cuda.empty_cache()
+                log(f"Sharded {component_name}. "
+                    f"(VRAM after: {torch.cuda.memory_allocated(local_rank)/1e9:.2f}GB)")
             else:
-                log(f"Skipping FSDP wrapping for {component_name}...")
+                log(f"i m wrapping for {component_name}...")
                 if hasattr(component, "to"):
                     component.to(f"cuda:{local_rank}")
                 else:
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -568,7 +568,8 @@ class xFuserModel(abc.ABC):
         if self.config.fully_shard_degree > 1:
             self._shard_model_with_fsdp()
         else:
-            self.pipe = self.pipe.to(f"cuda:{local_rank}")
+            if not (self.config.enable_model_cpu_offload or self.config.enable_sequential_cpu_offload):
+                self.pipe = self.pipe.to(f"cuda:{local_rank}")
             if self.config.use_fp4_gemms:
                 if _is_cuda():
                     self._setup_nvfp4_gemms(local_rank=local_rank)
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -627,7 +627,7 @@ class xFuserModel(abc.ABC):
                 log(f"Sharded {component_name}. "
                     f"(VRAM after: {torch.cuda.memory_allocated(local_rank)/1e9:.2f}GB)")
             else:
-                log(f"i m wrapping for {component_name}...")
+                log(f"Skipping FSDP wrapping for {component_name}...")
                 if hasattr(component, "to"):
                     component.to(f"cuda:{local_rank}")
                 else:
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -609,6 +609,7 @@ class xFuserModel(abc.ABC):
                 strategy = self.settings.fsdp_strategy[component_name]
                 wrap_attrs = strategy.get("wrap_attrs", [])
                 dtype = strategy.get("dtype", None)
+                offload_policy = strategy.get("offload_policy", None)
                 quantize_fn = self._build_fsdp_quantize_fn(component_name, wrap_attrs, fs_local_rank)
                 reshard_after_forward = self.config.reshard_after_forward
                 fsdp_object = shard_component(
@@ -616,6 +617,10 @@ class xFuserModel(abc.ABC):
                     quantize_fn=quantize_fn,
                     reshard_after_forward=reshard_after_forward,
                     memory_efficient_init=self.config.memory_efficient_sharding,
+                    offload_policy=offload_policy,
+                    # All ranks load from the same checkpoint so states are already
+                    # identical. Skip the GPU broadcast to avoid OOM on large encoders.
+                    sync_module_states=offload_policy != "cpu",
                 )
                 setattr(self.pipe, component_name, fsdp_object)
                 torch.cuda.empty_cache()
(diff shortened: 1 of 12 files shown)


xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -229,10 +247,24 @@ class xFuserFlux2Model(xFuserModel):
...
+                "offload_policy": "cpu",
+            },
         }
     )
 
+    def _post_load_and_state_initialization(self, input_args: dict) -> None:
+        super()._post_load_and_state_initialization(input_args)
+        te = self.pipe.text_encoder
+        # vision_tower is never called for FLUX.2 text-only embedding; keep on CPU.
+        if hasattr(te, "model") and hasattr(te.model, "vision_tower"):
```
avjves, line 275: In what cases is this used then? Just wondering why it's there. Is it used when also providing an input image?

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -229,10 +247,24 @@ class xFuserFlux2Model(xFuserModel):
         fsdp_strategy={
             "transformer": {
                 "wrap_attrs": ["transformer_blocks", "single_transformer_blocks"],
-            }
+            },
+            "text_encoder": {
+                "wrap_attrs": ["model.language_model.layers"],
+                "offload_policy": "cpu",
```
avjves, line 266: Does this affect the FSDP perf? Both adding TE sharding + cpu offloading
Sending a DM regarding this

xfuser/model_executor/models/runner_models/hunyuan.py
```diff
@@ -260,6 +291,17 @@ def __init__(self, config: xFuserArgs) -> None:
         self.settings.model_name = "tencent/HunyuanVideo-1.5"
         self.settings.output_name = "hunyuan_video_1_5_sparse"
         self.settings.valid_tasks = ["i2v"]
+        self.settings.fsdp_strategy = {
+            "transformer": {
+                "wrap_attrs": ["transformer_blocks", "single_transformer_blocks"],
+            },
+            "text_encoder": {
+                "wrap_attrs": ["layers"],
```
avjves: Should this be [] as well like the parent? Or do they have different encoders as well?

xfuser/model_executor/models/runner_models/flux.py
pds-amd: I've double checked this. This is safe for the current xDiT implementation.

The `vision_tower` is only used to convert a provided image to patch embeddings for the language model to "optimise" the provided prompt based on the image.

  1. xDiT `_run_pipe` calls `self.pipe(image=..., ...)`. We don't provide [caption_upsample_temperature](https://github.com/huggingface/diffusers/blob/33becabe52a7c76101227c9210a321089e5d0dd7/src/diffusers/pipelines/flux2/pipeline_flux2.py#L869) so `upsample_prompt()` is never called.                                                                                                                                                                                                                                                             
  2. image only goes to the [image processor path](https://github.com/huggingface/diffusers/blob/33becabe52a7c76101227c9210a321089e5d0dd7/src/diffusers/pipelines/flux2/pipeline_flux2.py#L889)                                                                                                                                                                                                                                                   
  3. So the `vision_tower` is never called
  
Right now this just occupies VRAM unnecessarily.

xfuser/model_executor/models/runner_models/flux.py
pds-amd: I did not observe a performance regressions by adding this.

xfuser/model_executor/models/runner_models/flux.py
avjves: Ah, makes sense. No problem, then :)

xfuser/model_executor/models/runner_models/hunyuan.py
pds-amd: Genuine bug. Fixed.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -623,6 +634,30 @@ def _shard_model_with_fsdp(self) -> None:
...
+        # Fix: give every nn.Module component a minimal _hf_hook so _execution_device
+        # continues past them, with cpu-offloaded components advertising cuda.
+        cpu_offloaded = {
+            name for name, s in self.settings.fsdp_strategy.items()
+            if s.get("offload_policy") == "cpu"
+        }
+        if cpu_offloaded:
+            cuda_device = f"cuda:{local_rank}"
+
+            class _ExecDeviceHook:
```
avjves, line 649: This part is a bit weird - why do we advertise device on CPU as being on GPU?

avjves: Assuming the need for this is somehow caused by FSDP?

xfuser/model_executor/models/runner_models/hunyuan.py
```diff
@@ -248,6 +260,7 @@ class xFuserHunyuanvideo15SparseModel(xFuserHunyuanvideo15Model):
         enable_slicing=True,
         enable_tiling=True,
         supports_sparse_attention_backends=True,
+        fully_shard_degree=True,
     )
```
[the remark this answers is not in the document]
pds-amd: Mutation gone, handled with class-level `settings` now.

xfuser/model_executor/models/runner_models/hunyuan.py
```diff
@@ -260,6 +290,31 @@ class xFuserHunyuanvideo15SparseModel(xFuserHunyuanvideo15Model):
         enable_slicing=True,
         enable_tiling=True,
         supports_sparse_attention_backends=True,
+        fully_shard_degree=True,
+    )
+    settings = ModelSettings(
+        model_name="tencent/HunyuanVideo-1.5",
```
avjves: I feel like there was some unfortunate architectural jankiness in the model runner that required these to be done in init, not directly as ModelSettings(). @ksikiric do you remember the reason?

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Yeah this is a sharp edge.

FSDP CPU offload doesn't install accelerate's `_hf_hook`, but diffusers' `_execution_device` walks pipeline components and returns `cpu`. This ensures the correct device is advertised and FSDP handles the CPU <-> GPU weight movement.

xfuser/model_executor/models/runner_models/hunyuan.py
pds-amd: I haven't managed to actually get end-to-end runs for ltx and hunyuan working past initial model load yet, so I think I will actually revert this fsdp stuff for those two for now until I have more time to look into those specifically.

ksikiric: Yes there was something here, cannot remember the details now.. let me double check. IIRC, since they inherit from Hunyuan1.5, their model settings dataclass would be overwritten by the models settings in Hunyuan 1.5 instead, but not sure..

Did you @pds-amd verify that the correct model name is used when running this Sparse model?

xfuser/model_executor/models/runner_models/base_model.py
avjves: I see - I guess we'll have to live with it 🫣

xfuser/model_executor/models/runner_models/hunyuan.py
pds-amd: I'm going to exclude it for now until I can properly test the impact of the change across multiple different GPU architectures.

ksikiric: For reference, the Sparse Hunyuan model does not have any sparse attention support on gfx1201, only gfx942 and gfx950.. And on NV, it is only supported on hopper. 

But since we call `super.__init__` it will call Hunyuan1.5 init function, and there it overrides the model_name, so you would need to override them again in the init of distilled/sparse models, and then it is really no need to specify their own model settings dataclass anyways, as you would need the override in both cases.

avjves (approved):
LGTM! No regression observed and now that LTX2/Hunyuan are excluded, no questions either :)

avjves (commented):
EDIT:
Lmao, how did I manage to leave a review to an old already merged PR 🫣 Please ignore.

PR #708, opened 2026-05-21
feat: enable VAE tiling and slicing across all runner models

pds-amd:
## Summary

Two small changes that expose existing diffusers VAE memory optimizations through xDiT's runner.

### 1. VAE tiling (`--enable_tiling`)

Adds `enable_tiling=True` to `ModelCapabilities` for all models that were missing it. Previously, passing `--enable_tiling` raised a validation error for these models even though their VAEs support it.

VAE tiling decodes in spatial tiles to reduce peak VRAM during decode. The benefit is resolution-dependent. At higher resolutions it becomes essential for low VRAM cards.

On AMD AI Pro R9700 32GB (gfx1201), FLUX.1-dev at 2048×2048:

| Config | s/iter | Peak VRAM |
|--------|--------|-----------|
| 1 GPU + compile + model offload, no tiling | ~252 | ~32.5 GB |
| 1 GPU + compile + model offload + tiling | 139.55 | 29.5 GB |

### 2. VAE slicing (`--enable_slicing`)

Adds `enable_slicing=True` to `ModelCapabilities` for all models that were missing it. Same issue. `--enable_slicing` raised a validation error despite the VAEs supporting it via `AutoencoderMixin`.

VAE slicing decodes one image at a time when `batch_size > 1`, reducing peak VRAM during batch inference. No effect at batch size 1.

### DIFF
--- a/xfuser/model_executor/models/runner_models/ltx.py
+++ b/xfuser/model_executor/models/runner_models/ltx.py
@@ -56,6 +56,7 @@ class xFuserLTX23VideoModel(xFuserModel):
     capabilities = ModelCapabilities(
         ulysses_degree=True,
         ring_degree=True,
+        enable_tiling=True,
     )
 
     _STG_SCALE = 1.0
@@ -222,6 +223,7 @@ class xFuserLTX2VideoModel(xFuserModel):
     capabilities = ModelCapabilities(
         ulysses_degree=True,
         ring_degree=True,
+        enable_tiling=True,
     )
 
     def _load_model(self) -> DiffusionPipeline:
--- a/xfuser/model_executor/models/runner_models/ltx.py
+++ b/xfuser/model_executor/models/runner_models/ltx.py
@@ -110,6 +110,11 @@ class xFuserLTX23VideoModel(xFuserModel):
 
         return pipe
 
+    def _enable_options(self) -> None:
+        super()._enable_options()
+        if self.config.enable_slicing:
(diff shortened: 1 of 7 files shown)


xfuser/model_executor/models/runner_models/ltx.py
```diff
@@ -108,6 +110,11 @@ def _load_model(self) -> DiffusionPipeline:
 
         return pipe
 
+    def _enable_options(self) -> None:
+        super()._enable_options()
+        if self.config.enable_slicing:
+            self.second_pipe.vae.enable_slicing()
```
avjves, line 116: Should this not enable tiling for the second pipe as well? Or is there a reason to not do it?

avjves:
Looks good, just one question and also currently there's a conflict to be resolved :)

xfuser/model_executor/models/runner_models/ltx.py
pds-amd: ltx.py:106. `second_pipe.vae.enable_tiling()` is hardcoded in _load_model, always fires regardless of config. I didn't want to change existing behavior. Should I move this to `_enable_options`?

avjves: Ah, would you look at that. Yeah, let's not change the existing behavior :)

avjves (approved):
LGTM!

PR #711, opened 2026-05-25
Aiter sage ring attention

ksikiric:
This pull request enhances support for ring parallelism in the AITER SAGE and SAGE V2 attention backends. It introduces runtime checks to ensure that required features are present in the underlying attention functions, and updates the attention call logic to use new capabilities when available.

**Ring parallelism and feature support checks:**

* Added runtime checks in `_check_if_backend_compatible_with_current_configuration` to ensure that, when using ring parallelism, the `fav3_sage_wrapper_func` (for AITER SAGE) and `fav3_sage_mxfp4_wrapper` (for AITER SAGE V2) support the `return_lse` parameter, and that AITER SAGE also supports the `smooth_k` parameter for correct LSE computation. If these are missing, a clear error message is raised instructing the user to update AITER.

**Attention backend function enhancements:**

* Added utility functions `_check_aiter_sage_has_return_lse` and `_check_aiter_sage_v2_has_return_lse` to programmatically check for the presence of the `return_lse` parameter in the respective attention functions.
* In both `_aiter_sage_attn_call` and `_aiter_sage_v2_attn_call`, updated logic to pass `return_lse=True` and handle the returned `softmax_lse` only if the underlying function supports it and ring parallelism is enabled; otherwise, defaults to previous behavior.

**Imports and dependencies:**

* Imported `get_ring_parallel_world_size` to enable ring parallelism checks in the attention backend.

### DIFF
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -53,13 +53,14 @@ def _check_aiter_fp8_has_descale():
         AITER_FP8_HAS_DESCALE = False
     return AITER_FP8_HAS_DESCALE
 
-def _check_aiter_sage_has_return_lse():
+def _check_aiter_sage_supports_ring():
     try:
-        return inspect.signature(fav3_sage_wrapper_func).parameters.get("return_lse") is not None
+        parameters = inspect.signature(fav3_sage_wrapper_func).parameters
+        return "return_lse" in parameters and "smooth_k" in parameters
     except (NameError, ImportError, AttributeError, TypeError):
         return False
 
-def _check_aiter_sage_v2_has_return_lse():
+def _check_aiter_sage_v2_supports_ring():
     try:
         return inspect.signature(fav3_sage_mxfp4_wrapper).parameters.get("return_lse") is not None
     except (NameError, ImportError, AttributeError, TypeError):
@@ -336,8 +337,8 @@ if env_info["has_aiter"]:
     AITER_FP8_STATIC_SCALE_WITH_DESCALE, AITER_FP8_STATIC_SCALE_NO_DESCALE, AITER_SAGE_V2_BLOCK_R = _setup_aiter_environment_variables()
     AITER_HAS_ROUND_MODE, HOW_V3_BF16_CVT = _check_aiter_round_mode()
     AITER_FP8_HAS_DESCALE = _check_aiter_fp8_has_descale()
-    AITER_SAGE_HAS_RETURN_LSE = _check_aiter_sage_has_return_lse()
-    AITER_SAGE_V2_HAS_RETURN_LSE = _check_aiter_sage_v2_has_return_lse()
+    AITER_SAGE_SUPPORTS_RING = _check_aiter_sage_supports_ring()
+    AITER_SAGE_V2_SUPPORTS_RING = _check_aiter_sage_v2_supports_ring()
     HADAMARD_MATRIX = _aiter_sage_v2_hadamard_matrix(AITER_SAGE_V2_BLOCK_R)
     _TRITON_SSTA_BLOCK_SIZE = 128
     
@@ -771,8 +772,8 @@ def npu_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwarg
 @register_attention_function(AttentionBackendType.AITER_SAGE)
 def _aiter_sage_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=None):
     # Pass layout="bhsd" to avoid permutation
-    if AITER_SAGE_HAS_RETURN_LSE and get_ring_parallel_world_size() > 1:
-        attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd", return_lse=True)
+    if AITER_SAGE_SUPPORTS_RING and get_ring_parallel_world_size() > 1:
+        attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd", return_lse=True, smooth_k=True)
         output, softmax_lse = attn_fn(query, key, value)
     else:
         attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd")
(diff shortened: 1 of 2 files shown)


xfuser/core/distributed/attention_backend.py
```diff
@@ -757,8 +771,13 @@ def npu_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwarg
 @register_attention_function(AttentionBackendType.AITER_SAGE)
 def _aiter_sage_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=None):
     # Pass layout="bhsd" to avoid permutation
-    attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd", return_lse=True)
-    output, softmax_lse = attn_fn(query, key, value)
+    if AITER_SAGE_HAS_RETURN_LSE and get_ring_parallel_world_size() > 1:
+        attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd", return_lse=True)
+        output, softmax_lse = attn_fn(query, key, value)
```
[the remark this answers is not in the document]
ksikiric: It defaults to True, will add comment about this

xfuser/core/distributed/runtime_state.py
```diff
@@ -214,7 +214,36 @@ def _check_if_backend_compatible_with_current_configuration(self, attention_back
...
+                    except ImportError:
+                        raise RuntimeError("AITER Sage V2 attention is not available, please update AITER") from None
+                    try:
+                        has_lse = inspect.signature(fav3_sage_mxfp4_wrapper).parameters.get("return_lse") is not None
+                    except (AttributeError, TypeError):
+                        has_lse = False
+                    if not has_lse:
+                        raise RuntimeError("AITER Sage V2 attention does not support return_lse, please update AITER")
+                else:
+                    raise RuntimeError("Selected attention backend does not support ring parallelism.")
```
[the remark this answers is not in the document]
ksikiric: All runtime config validations are checked here, and will raise an error if wrong. This is done before the actual run, so no effect on performance. I will address the Code duplication though.

xfuser/core/distributed/attention_backend.py
```diff
@@ -53,6 +53,18 @@ def _check_aiter_fp8_has_descale():
         AITER_FP8_HAS_DESCALE = False
     return AITER_FP8_HAS_DESCALE
 
+def _check_aiter_sage_has_return_lse():
+    try:
+        return inspect.signature(fav3_sage_wrapper_func).parameters.get("return_lse") is not None
+    except (NameError, ImportError, AttributeError, TypeError):
+        return False
```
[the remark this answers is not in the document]
ksikiric: True, but we will never actually be able to run if smooth_k is not present, since the validation in runtime_state.py will throw an error in that case. The fix for the lse computation was added with the smooth_k, so ring attention will only return correct videos with this version of AITER, but only return_lse is strictly needed to return the softmax_lse.

I will add check for smooth_k here though for consistency.

xfuser/core/distributed/attention_backend.py
```diff
@@ -757,8 +772,13 @@ def npu_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwarg
 @register_attention_function(AttentionBackendType.AITER_SAGE)
 def _aiter_sage_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=None):
     # Pass layout="bhsd" to avoid permutation
-    attn_fn = functools.partial(fav3_sage_wrapper_func, layout="bhsd", return_lse=True)
-    output, softmax_lse = attn_fn(query, key, value)
+    if AITER_SAGE_SUPPORTS_RING and get_ring_parallel_world_size() > 1:
```
avjves, line 775: Does this slow things down? 
Could we just return it by default?

xfuser/core/distributed/runtime_state.py
```diff
@@ -214,7 +215,40 @@ def _check_if_backend_compatible_with_current_configuration(self, attention_back
                                  AttentionBackendType.AITER_SPARSE_SAGE_V2,
                                  AttentionBackendType.FLEX_BLOCK_ATTN]:
             if self.parallel_config.ring_degree > 1:
-                raise RuntimeError("Selected attention backend does not support ring parallelism.")
+                # Ring parallelism merges per-rank attention outputs via LSE, so
```
avjves, line 218: This function is getting a bit out of hand 😬 
I wonder if we should include the required AITER in the setup (on ROCm) and could just skip these import tests 🤔

avjves: Maybe that's outside the scope of this PR. A cleanup PR should maybe be done, eventually 😄

xfuser/core/distributed/attention_backend.py
ksikiric: Should not slow anything down, both AITER_SAGE_SUPPORTS_RING and get_ring_parallel_world_size() are constants during runtime. 

Even if we return it by default we would still need a check to see if sage supports ring, or did I misunderstand you?

xfuser/core/distributed/runtime_state.py
ksikiric: I can take a look intot cleaning all this up 😄

xfuser/core/distributed/attention_backend.py
avjves: Basically I'm wondering if there's a need to check that if it supports it. We add a lot of compatiblity checks for these, but I don't think anyone is using these with an older AITER? 🤔 

I suppose some recommended / auto installed AITER version on ROCm would solve these? Again, maybe not in the scope of this PR, just wanted your opinion :)

And there's no slowdown to return the lse? If we knew that the AITER was up-to-date

avjves (approved):

avjves:
I'll merge this - lets make a follow up PR later to clean up things a bit and maybe look into getting rid of AITER "version checks"

PR #712, opened 2026-05-26
feat: FP8 block-scale GEMMs via AITER for RDNA4 (gfx1200+)

pds-amd:
## Summary

Adds true FP8 GEMM execution for transformer linear layers on AMD RDNA4 GPUs (gfx1200+, e.g. AI Pro R9700), using AITER's `gemm_a8w8_blockscale` Triton kernel which hits native FP8 WMMA hardware.

### Why not torchao?

torchao's `Float8DynamicActivationFloat8WeightConfig` is currently gated to CUDA>=9.0 and MI300+. Test patching this path resulted in very slow inference. It dispatches through `torch._scaled_mm` to hipBLASLt. On ROCm 7.2.3, gfx1201 is not in hipBLASLt's FP8 whitelist and the GEMM silently falls back to FP32. AITER's Triton kernel auto-compiles to native `v_wmma_f32_16x16x16` instructions and is confirmed working on gfx1201.

### Implementation

**`xFuserFP8BlockScaleLinear`** is a drop-in `nn.Module` replacement for `nn.Linear`:
- Weights pre-quantized at load time to FP8 e4m3fn with block-128 scaling; stored as persistent FSDP2-compatible buffers
- Activations quantized per-forward using manual block-128 quant (reshape -> amax -> scale -> cast); placed **outside** the custom_op boundary so Inductor can fuse the quant ops into ~2 kernels
- The bare `aiter.gemm_a8w8_blockscale` call is wrapped in `torch.library.custom_op` + `register_fake` for `torch.compile` compatibility
- Bias added after the GEMM (kernel has no native bias support)

**Gate**: `_use_aiter_fp8_path()` restricts to gfx1200+ (RDNA4 only).

### Models enabled

All models available through the runner: FLUX.1-dev, FLUX.1-Kontext, FLUX.2, FLUX.2-klein, HunyuanVideo, HunyuanVideo-1.5, LTX-2, Qwen-Image, Qwen-Image-Edit, StableDiffusion-3.5, Wan2.1, Wan2.2, Z-Image.

The `fp8_gemm_module_list` in each model's `ModelSettings` controls which submodules are quantized (e.g. `transformer.transformer_blocks`).

### Usage

```bash
torchrun ... --use_fp8_gemms
```

### DIFF
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -11,7 +11,7 @@ from xfuser.envs import _is_cuda, _is_hip
 logger = logging.getLogger(__name__)
 
 
-def _use_aiter_fp8_path() -> bool:
+def _use_aiter_fp8_rdna4() -> bool:
     """True on ROCm gfx1200 (Navi 44) or gfx1201 (Navi 48) with AITER present."""
     if not _is_hip():
         return False
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -323,9 +323,7 @@ def quantize_linear_layers_to_fp8_blockscale(
                 device=module.weight.device,
                 dtype=module.weight.dtype,
             )
-            fp8_layer.load_and_quantize_weights(module.weight, module.bias)
-            if device is not None:
-                fp8_layer.to(device)
+            fp8_layer.load_and_quantize_weights(module.weight, module.bias, device=device)
             setattr(model, name, fp8_layer)
         elif next(module.children(), None) is not None:
             quantize_linear_layers_to_fp8_blockscale(module, device=device)
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -325,7 +325,7 @@ def quantize_linear_layers_to_fp8_blockscale(
                 device=weight.device,
                 dtype=weight.dtype,
             )
-            # Free BF16 weight before FP32 quant copy to halve peak memory per layer
+            # Free BF16 weight before quantization to avoid holding two copies on GPU
             module.weight = None
             if module.bias is not None:
                 module.bias = None
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -316,14 +316,21 @@ def quantize_linear_layers_to_fp8_blockscale(
 
     for name, module in list(model.named_children()):
         if isinstance(module, torch.nn.Linear):
+            weight = module.weight.data
+            bias = module.bias.data if module.bias is not None else None
             fp8_layer = xFuserFP8BlockScaleLinear(
                 module.in_features,
                 module.out_features,
-                bias=(module.bias is not None),
-                device=module.weight.device,
-                dtype=module.weight.dtype,
+                bias=(bias is not None),
+                device=weight.device,
+                dtype=weight.dtype,
             )
-            fp8_layer.load_and_quantize_weights(module.weight, module.bias, device=device)
+            # Free BF16 weight before FP32 quant copy to halve peak memory per layer
+            module.weight = None
+            if module.bias is not None:
+                module.bias = None
+            fp8_layer.load_and_quantize_weights(weight, bias, device=device)
+            del weight, bias
             setattr(model, name, fp8_layer)
         elif next(module.children(), None) is not None:
             quantize_linear_layers_to_fp8_blockscale(module, device=device)
--- a/xfuser/core/utils/runner_utils.py
+++ b/xfuser/core/utils/runner_utils.py
@@ -6,9 +6,32 @@ import functools
 import numpy as np
 from PIL.Image import Image
 from typing import Callable, Optional
-from xfuser.envs import _is_cuda
+from xfuser.envs import _is_cuda, _is_hip
 
 logger = logging.getLogger(__name__)
+
+
+def _use_aiter_fp8_path() -> bool:
+    """True on ROCm RDNA4+ (gfx1200+) with AITER present.
+
+    Restricted to gfx1200+ (RDNA4: R9700, etc.) where our AITER block-128
+    path has been validated and torchao/hipBLASLt FP8 is unavailable (ROCm 7.2.3).
+    CDNA3+ (MI300/MI350, gfx940-999) has battle-tested torchao FP8 and falls
+    through to that path instead.
+    """
+    if not _is_hip():
+        return False
+    try:
+        import aiter  # noqa: F401
+    except ImportError:
+        return False
+    import re
+    arch = torch.cuda.get_device_properties(0).gcnArchName.split(":")[0]
+    m = re.match(r"gfx(\d+)", arch)
+    if not m:
+        return False
+    n = int(m.group(1))
+    return n >= 1200
 
 def log(message: str, debug=False, log_from_all_processes: bool = False) -> None:
     """Log message. By default, only from the last process to avoid duplicates."""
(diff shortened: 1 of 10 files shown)


xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@torch.library.custom_op("mylib::fp8_blockscale_gemm", mutates_args=())
+def _fp8_blockscale_gemm(
+    a_q: torch.Tensor,
+    a_scale: torch.Tensor,
+    w_fp8: torch.Tensor,
+    w_scale: torch.Tensor,
+) -> torch.Tensor:
+    return aiter.gemm_a8w8_blockscale(a_q, w_fp8, a_scale, w_scale)
```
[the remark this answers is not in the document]
pds-amd: Intentional. This mirrors xFuserMXFP4Linear which also uses `mylib`.

xfuser/core/utils/runner_utils.py
```diff
@@ -280,6 +303,42 @@ def quantize_linear_layers_to_fp8(module_or_module_list_to_quantize: torch.nn.Mo
...
+                if fp8_layer.weight_fp8 is not None:
+                    fp8_layer.weight_fp8 = fp8_layer.weight_fp8.to(device)
+                if fp8_layer.weight_scale is not None:
+                    fp8_layer.weight_scale = fp8_layer.weight_scale.to(device)
+                if fp8_layer.bias is not None:
+                    fp8_layer.bias.data = fp8_layer.bias.data.to(device)
+            setattr(model, name, fp8_layer)
+        elif len(list(module.children())) > 0:
+            full_name = f"{parent_name}.{name}" if parent_name else name
+            quantize_linear_layers_to_fp8_blockscale(module, full_name, device=device)
```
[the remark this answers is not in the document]
pds-amd: Fixed.

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -0,0 +1,145 @@
...
+        x_scale = (x_amax / _FP8_MAX).float()
+        x_q = (x_blocks / x_amax.unsqueeze(-1) * _FP8_MAX).clamp(-_FP8_MAX, _FP8_MAX)
+        x_q = x_q.to(torch.float8_e4m3fn).reshape(M, K_pad)[:, :K].contiguous()
+
+        output = torch.ops.mylib.fp8_blockscale_gemm(
+            x_q, x_scale, self.weight_fp8, self.weight_scale,
+        )
+        if self.bias is not None:
+            output = output + self.bias
+        return output.view(*original_shape[:-1], self.out_features)
```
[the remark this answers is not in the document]
pds-amd: Fixed.

xfuser/core/utils/runner_utils.py
```diff
@@ -6,10 +6,27 @@
...
+        import aiter  # noqa: F401
+    except ImportError:
+        return False
+    import re
+    arch = torch.cuda.get_device_properties(0).gcnArchName.split(":")[0]
+    m = re.match(r"gfx(\d+)", arch)
+    if not m:
+        return False
+    n = int(m.group(1))
+    return n >= 1200
```
nsakkine: Feels slightly too generic and could clash with future non-RDNA4 targets? Can we be more specific?

pds-amd: Locked it to gfx1200 and gfx1201. Better?

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -0,0 +1,150 @@
...
+    ) -> None:
+        self._quantize_weights(weight, device=device)
+        if bias is not None and self.bias is not None:
+            target = device if device is not None else bias.device
+            self.bias = torch.nn.Parameter(bias.to(device=target, dtype=bias.dtype).detach())
+
+    def _quantize_weights(self, weight: torch.Tensor, device: Optional[torch.device] = None) -> None:
+        N, K = weight.shape
+        n_blocks = math.ceil(N / _FP8_BLOCK)
+        k_blocks = math.ceil(K / _FP8_BLOCK)
```
nsakkine: Nitpicky but `(N + _FP8_BLOCK - 1) // _FP8_BLOCK` would avoid floating-point representation

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -0,0 +1,150 @@
...
+        x_amax = x_blocks.abs().amax(dim=-1).clamp(min=1e-12)  # [M, k_blocks]
+        x_scale = (x_amax.float() / _FP8_MAX)
+        x_q = (x_blocks / x_amax.unsqueeze(-1) * _FP8_MAX).clamp(-_FP8_MAX, _FP8_MAX)
+        x_q = x_q.to(torch.float8_e4m3fn).reshape(M, K_pad)
+        if needs_pad:
+            x_q = x_q[:, :K].contiguous()
+
+        output = torch.ops.mylib.fp8_blockscale_gemm(
+            x_q, x_scale, self.weight_fp8, self.weight_scale,
+        ).to(input.dtype)
```
nsakkine, line 141: Does pre-shuffling play a role at all?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -24,9 +24,11 @@
     log,
     load_dataset_prompts,
     quantize_linear_layers_to_fp8,
+    quantize_linear_layers_to_fp8_blockscale,
     quantize_linear_layers_to_fp4,
     quantize_linear_layers_to_nvfp4,
     convert_model_convs_to_channels_last,
+    _use_aiter_fp8_path,
```
nsakkine: More specific name if this is meant for RDNA4 only?

xfuser/core/utils/runner_utils.py
nsakkine: I believe this is better, thanks 🙏

xfuser/core/utils/runner_utils.py
```diff
@@ -6,10 +6,27 @@
...
+    except ImportError:
+        return False
+    import re
+    arch = torch.cuda.get_device_properties(0).gcnArchName.split(":")[0]
+    m = re.match(r"gfx(\d+)", arch)
+    if not m:
+        return False
+    n = int(m.group(1))
+    return n in {1200, 1201}
+
```
lauri9, line 23: Just a drive-by comment. I think your other PR introduced `_is_rdna4()`, could we use it here to replace the custom regexp logic?

For example:
```python
# enable when AITER present on RDNA4
try:
    import aiter
except ImportError:
   return False
return _is_rdna4()
```

pds-amd: Good catch! Will centralize.

xfuser/model_executor/layers/fp8_linear.py
pds-amd: I don't believe so... [It looks to me like it won't apply in this case yet.](https://github.com/ROCm/aiter/blob/28dc8c411455174400c447a6ad644190139110d2/aiter/ops/gemm_op_a8w8.py#L698)

xfuser/model_executor/layers/fp8_linear.py
pds-amd: Changed. Better, thanks!

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Changed

nsakkine (approved):
Thank you, looks good 👍

avjves (approved):

PR #713, opened 2026-05-28
Solve Hunyuan+AITER Sage regression with FX pass solution

ksikiric:
## What

Defeats an Inductor scheduler cost-model bug that regresses Hunyuan video (HV and HV1.5) on Ulysses SP with the AITER Sage attention backend when the encoder length isn't divisible by the SP world size (asymmetric joint attention path).

In that configuration, USP cats the encoder K/V onto the latent K/V *after* the all-to-all, then feeds the result into AITER's `sage_quant`. Sage exposes its quantization-scale computations as eager Python reductions (`v.abs().amax(dim=2)`, `k.mean(dim=2)`). Inductor materializes the cat once for the sage Triton kernel's `V_Input`/`K_Input`, but for the reduction consumers it chooses to **re-derive** the cat from sources, fusing the entire post-A2A view chain into the reduction. The cost model doesn't price stride patterns, so it doesn't see that re-derivation is a strided gather, producing a ~9.5 ms/call kernel (`triton_red_fused__unsafe_view_abs_amax_cat_clone_permute_slice_split_with_sizes_transpose_view_*`).

## How

A custom Inductor post-grad FX pass (`fix_sage_joint_cat_fusion`) that:

1. Defines an opaque marker custom op `xfuser::_realize_for_inductor` (semantically `x.clone()`).
2. Walks each compiled subgraph for `aten.cat.default` nodes that have ≥2 users with at least one being `aten.abs.default` or `aten.mean.dim` — the unique FX signature of the bad pattern.
3. Inserts the marker between the cat and its reduction consumer and rewires the reduction to read from the marker.

Because the marker is a custom op, Inductor cannot inline through it. The reduction reads coalesced from the materialized cat buffer instead of strided-gathering over the view chain.

## Scope

- Pass is installed only from `xFuserHunyuanvideoModel._compile_model` and `xFuserHunyuanvideo15Model._compile_model` → other models (Flux, Wan, etc.) never register the hook.
- Pattern only matches the specific cat→reduction shape produced by sage's eager scales → Hunyuan + non-sage backends and Hunyuan symmetric joint compile to no-op.
- No change to model code, USP, AITER, or attention backends.

## Files

- **Added** `xfuser/compile/__init__.py`, `xfuser/compile/inductor_passes.py` — custom op + pass + `install()`.
- **Modified** `xfuser/model_executor/models/runner_models/hunyuan.py` — call `install_inductor_passes()` at top of `_compile_model` for HV and HV1.5.

## Reversibility

Comment out `install_inductor_passes()` in `_compile_model` and clear `/tmp/torchinductor_*` → reverts to default Inductor scheduling. Two-line revert.

## Follow-ups

- File an upstream PyTorch issue with a minimal repro so the cost-model bug can be fixed and this pass eventually retired.

### DIFF
--- a/xfuser/compile/__init__.py
+++ b/xfuser/compile/__init__.py
@@ -1,0 +1,3 @@
+from xfuser.compile.inductor_passes import install as install_inductor_passes
+
+__all__ = ["install_inductor_passes"]
--- a/xfuser/compile/inductor_passes.py
+++ b/xfuser/compile/inductor_passes.py
@@ -11,11 +11,6 @@ Currently provides:
 The pass is pattern-matched on the FX graph and is a no-op when the bad
 pattern is not present, so it is safe to install globally; non-matching
 configurations are untouched.
-
-This pass does NOT introduce a graph break. ``torch.library.custom_op`` is
-an Inductor opacity barrier (the op is treated as opaque for fusion purposes),
-not a Dynamo bailout: the FX graph stays connected as a single compiled
-subgraph and the AOT segment count is unchanged.
 """
 
 from __future__ import annotations
--- a/xfuser/compile/inductor_passes.py
+++ b/xfuser/compile/inductor_passes.py
@@ -150,19 +150,17 @@ def fix_sage_joint_cat_fusion(graph: fx.Graph) -> fx.Graph:
 
         num_matches += 1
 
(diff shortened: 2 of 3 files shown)


jcaraban (approved):

avjves (approved):
LGTM!

PR #714, opened 2026-05-29
fix: restore FlyDSL non-causal threshold fallback

pds-amd:
I over-corrected by falling back to `sdpa_flash` for all non-causal attention.

The kernel raises only when `seq_len` padding to 128 exceeds 0.5%, not for all non-causal attention. Padding is handled internally when below the threshold.

Tested by performing a full sweep across models and workloads. Speed gain is net positive as this reduces the number of fallback calls to `sdpa_flash`.

### DIFF
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -971,10 +971,15 @@ def _aiter_sparge_v2_attn_call(query, key, value, dropout_p, is_causal, attentio
 
 @register_attention_function(AttentionBackendType.AITER_FLYDSL)
 def _aiter_flydsl_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=None):
-    # This kernel only supports causal self-attention; fall back for cross-attention
-    # and non-causal attention.
-    if not is_causal or query.shape[2] != key.shape[2]:
+    # Cross-attention is unsupported. For non-causal self-attention, this kernel raises
+    # when the seq_len padding ratio to align to 128 exceeds 0.5% (same check as the kernel).
+    if query.shape[2] != key.shape[2]:
         return _sdpa_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs)
+    if not is_causal:
+        seq_len = query.shape[2]
+        pad = (-seq_len) % 128
+        if pad > 0 and pad / (seq_len + pad) > 0.005:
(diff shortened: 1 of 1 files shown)


xfuser/core/distributed/attention_backend.py
```diff
@@ -971,10 +971,15 @@ def _aiter_sparge_v2_attn_call(query, key, value, dropout_p, is_causal, attentio
...
-    if not is_causal or query.shape[2] != key.shape[2]:
+    # Cross-attention is unsupported. For non-causal self-attention, this kernel raises
+    # when the seq_len padding ratio to align to 128 exceeds 0.5% (same check as the kernel).
+    if query.shape[2] != key.shape[2]:
         return _sdpa_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs)
+    if not is_causal:
+        seq_len = query.shape[2]
+        pad = (-seq_len) % 128
+        if pad > 0 and pad / (seq_len + pad) > 0.005:
+            return _sdpa_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs)
```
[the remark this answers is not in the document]
pds-amd: Changed.

xfuser/core/distributed/attention_backend.py
```diff
@@ -971,10 +971,15 @@ def _aiter_sparge_v2_attn_call(query, key, value, dropout_p, is_causal, attentio
...
-    # and non-causal attention.
-    if not is_causal or query.shape[2] != key.shape[2]:
+    # Cross-attention is unsupported. For non-causal self-attention, this kernel raises
+    # when the seq_len padding ratio to align to 128 exceeds 0.5% (same check as the kernel).
+    if query.shape[2] != key.shape[2]:
         return _sdpa_flash_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs)
+    if not is_causal:
+        seq_len = query.shape[2]
+        pad = (-seq_len) % 128
+        if pad > 0 and 199 * pad > seq_len:
```
avjves, line 982: I see that this is now slightly better per Gemini's suggestion, but that 199 feels like a magic number on a first glance - could do with a comment explaining it. I for sure would be confused if I read it without prior knowledge

pds-amd: I had to manually verify it because I didn't trust it entirely at first myself. Comment added.

avjves (approved):

PR #719, opened 2026-06-01
feat: FBCache step-caching support for FLUX.2 models

nehaprakriya:
## Summary

Add First-Block Cache (FBCache) support for FLUX.2-dev and FLUX.2-klein models.

FBCache skips redundant transformer block computation across denoising steps by running only the first transformer block and comparing its output residual to the cached value from the previous step. If the relative L1 distance is below the threshold, the remaining 55 blocks are skipped entirely and the cached output is reused.

**Results at 2048x2048, 8-GPU Ulysses, BF16, 50 denoising steps:**
- 3.74x throughput improvement (0.1035 → 0.3876 img/s)
- Quality preserved: SSIM=0.9579, cosine similarity=0.998, LPIPS=0.0275

## Changes

- New FLUX.2-specific FBCache adapter (`flux2.py`) handling dual-stream and single-stream block signatures
- Register Flux2Transformer2DModel in FBCache adapter registry
- FBCache-aware `torch.compile` mode selection (`mode="default"` when FBCache active — CUDA graphs are incompatible with cross-step tensor caching)
- Add `set_input_parameters` call in FLUX.2 `_run_pipe` for correct USP distributed state
- Add `_compile_model` override for both FLUX.2-dev and FLUX.2-klein

## Usage

```bash
export XDIT_USE_FBCACHE=1
export XDIT_FBCACHE_THRESH=0.12  # lower = more conservative, higher = faster but lower quality
export XDIT_USE_PARALLEL_VAE=1

### DIFF
--- a/xfuser/model_executor/models/runner_models/flux.py
+++ b/xfuser/model_executor/models/runner_models/flux.py
@@ -1,4 +1,3 @@
-import os
 import torch
 from diffusers import FluxPipeline, FluxKontextPipeline, Flux2Pipeline, Flux2KleinPipeline
 from diffusers.pipelines.pipeline_utils import DiffusionPipeline
@@ -243,6 +242,7 @@ class xFuserFlux2Model(xFuserModel):
         enable_tiling=True,
         enable_slicing=True,
         use_parallel_vae=True,
+        use_fbcache=True,
     )
     default_input_values = DefaultInputValues(
         height=1024,
@@ -274,15 +274,12 @@ class xFuserFlux2Model(xFuserModel):
         if self.config.use_parallel_vae:
             _setup_parallel_vae(self.pipe.vae)
 
-        use_fbcache = (
-            getattr(self.config, "use_fbcache", False)
-            or os.environ.get("XDIT_USE_FBCACHE", "0") == "1"
-        )
-        if use_fbcache:
+        if self.config.use_fbcache:
             from xfuser.model_executor.cache.diffusers_adapters.flux2 import (
                 apply_cache_on_transformer as apply_flux2_cache,
             )
-            rel_l1_thresh = float(os.environ.get("XDIT_FBCACHE_THRESH", "0.12"))
+            from xfuser.envs import XDIT_FBCACHE_THRESH
+            rel_l1_thresh = float(XDIT_FBCACHE_THRESH) if XDIT_FBCACHE_THRESH else self.settings.fbcache_thresh
             num_steps = int(input_args.get("num_inference_steps", 50))
             apply_flux2_cache(
                 self.pipe.transformer,
@@ -303,8 +300,7 @@ class xFuserFlux2Model(xFuserModel):
         because FBCache's cross-step tensor caching is incompatible with CUDA
         graphs (buffer aliasing on dynamo guard changes).
         """
-        use_fbcache = os.environ.get("XDIT_USE_FBCACHE", "0") == "1"
-        compile_mode = "default" if use_fbcache else "reduce-overhead"
+        compile_mode = "default" if self.config.use_fbcache else "reduce-overhead"
         torch._inductor.config.reorder_for_compute_comm_overlap = True
         self.pipe.transformer = torch.compile(
             self.pipe.transformer,
@@ -339,13 +335,6 @@ class xFuserFlux2Model(xFuserModel):
         return input_args
 
     def _run_pipe(self, input_args: dict) -> DiffusionOutput:
-        batch_size = self.config.batch_size if self.config.batch_size else 1
-        get_runtime_state().set_input_parameters(
-            batch_size=batch_size,
-            num_inference_steps=input_args["num_inference_steps"],
-            max_condition_sequence_length=input_args["max_sequence_length"],
-            split_text_embed_in_sp=get_pipeline_parallel_world_size() == 1,
-        )
         output = self.pipe(
             height=input_args["height"],
             width=input_args["width"],
@@ -371,6 +360,7 @@ class xFuserFlux2Klein9BModel(xFuserModel):
         enable_slicing=True,
         use_parallel_vae=True,
         fully_shard_degree=True,
+        use_fbcache=True,
     )
 
     default_input_values = DefaultInputValues(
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: xfuser/model_executor/models/runner_models/flux.py [hunk 1]
```diff
@@ -273,6 +274,45 @@ def _post_load_and_state_initialization(self, input_args: dict) -> None:
         if self.config.use_parallel_vae:
             _setup_parallel_vae(self.pipe.vae)
 
+        use_fbcache = (
```
avjves: We try to avoid having envionment variables for generic features such as FBCache. The better option would be to use the pre-existing `use_fbcache` cli argument to enable it and check it from `self.config` instead of `os.environ`. The fbcache is probably not wired in for the model runner args, so that should be added also in the `args.py` `add_runner_args`.

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -273,6 +274,45 @@ def _post_load_and_state_initialization(self, input_args: dict) -> None:
...
+            getattr(self.config, "use_fbcache", False)
+            or os.environ.get("XDIT_USE_FBCACHE", "0") == "1"
+        )
+        if use_fbcache:
+            from xfuser.model_executor.cache.diffusers_adapters.flux2 import (
+                apply_cache_on_transformer as apply_flux2_cache,
+            )
+            rel_l1_thresh = float(os.environ.get("XDIT_FBCACHE_THRESH", "0.12"))
```
avjves: Is the default value for this (0.12) specific to FLUX? We should add this to ModelSettings(), so each model can define this in the future.

If it needs to be modified for the model, then this setting is required. The `XDIT_FBCACHE_THRESH` doesn't need to be a CLI argument necessarily, as it's an optional change for the generic feature. So environment variable is fine, but it should introduced in `envs.py` and imported from there:
https://github.com/xdit-project/xDiT/blob/dd2aead53cfe1a480a330f4b8993a040dee74148/xfuser/envs.py#L49

xfuser/model_executor/models/runner_models/flux.py
avjves: Would be good to add a check in the `base_model.py` `_validate_config` to also then validate that the model supports fbcache and the arg isn't just silently ignored. A new line in the `ModelCapabilities` for this should be added. I suppose currently only the FLUX.2 supports it?

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -273,6 +274,45 @@ def _post_load_and_state_initialization(self, input_args: dict) -> None:
...
+            )
+
+    def _compile_model(self, input_args: dict) -> None:
+        """Compile the model using torch.compile.
+
+        When FBCache is enabled, uses mode="default" instead of "reduce-overhead"
+        because FBCache's cross-step tensor caching is incompatible with CUDA
+        graphs (buffer aliasing on dynamo guard changes).
+        """
+        use_fbcache = os.environ.get("XDIT_USE_FBCACHE", "0") == "1"
```
avjves: Here too the check should come from `self.config` :)

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -299,6 +339,13 @@ def _preprocess_args_images(self, input_args: dict) -> dict:
         return input_args
 
     def _run_pipe(self, input_args: dict) -> DiffusionOutput:
+        batch_size = self.config.batch_size if self.config.batch_size else 1
+        get_runtime_state().set_input_parameters(
```
avjves: This `set_input_parameters` call is not used by FLUX.2 - only some older models still require it. This can be removed for simplicity.

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -352,6 +399,24 @@ def _post_load_and_state_initialization(self, input_args: dict) -> None:
...
             _setup_parallel_vae(self.pipe.vae)
 
+    def _compile_model(self, input_args: dict) -> None:
+        """Compile the model using torch.compile.
+
+        When FBCache is enabled, uses mode="default" instead of "reduce-overhead"
+        because FBCache's cross-step tensor caching is incompatible with CUDA
+        graphs (buffer aliasing on dynamo guard changes).
+        """
+        use_fbcache = os.environ.get("XDIT_USE_FBCACHE", "0") == "1"
```
avjves: Same as above

avjves (commented):
Thanks for this PR!
I left a few comments that should still be addressed first :)

xfuser/model_executor/models/runner_models/flux.py
nehaprakriya: Thanks for the feedback: 1) Switched to `self.config.use_fbcache` throughout. 2) Added `use_fbcache: bool = False` to `ModelCapabilities` in `base_model.py`. The existing `_validate_config` loop now automatically rejects `--use_fbcache` if the model doesn't declare support. Currently only FLUX.2-dev and FLUX.2-klein set `use_fbcache=True` in their capabilities.

xfuser/model_executor/models/runner_models/flux.py
nehaprakriya: Yes, 0.12 was validated specifically for FLUX.2. Other models will likely need different values. Added `fbcache_thresh: float = 0.12` to `ModelSettings` so each model can override its default. The env var `XDIT_FBCACHE_THRESH` is now registered in `envs.py` and imported from there (it takes priority over the per-model default when set).

xfuser/model_executor/models/runner_models/flux.py
nehaprakriya: Fixed - reads from the 'self.config.use_fbcache' now.

xfuser/model_executor/models/runner_models/flux.py
nehaprakriya: Thanks, I have removed it.

xfuser/model_executor/models/runner_models/flux.py
nehaprakriya: Fixed - same change applied here.

nehaprakriya:
Thanks for the review @avjves ! I have addressed it in the second commit. Please let me know if further changes are required.

avjves (approved):
Looks good now! I pushed one small fix (added the fbcache cli arg to runner). Thanks :)

PR #721, opened 2026-06-08
Flex block sparge

ksikiric:
The `FLEX_BLOCK_SPARGE` work adds a new flex-attention sparse backend that reuses the existing sparge mask/reorder machinery: in `attention_backend.py` it registers the backend and shares the `_build_sparge_block_mask` path with aiter_sparge (sparge setup → mean-sim dynamic mask OR'd with the static block-neighbor mask → padded-KV-block masking), then runs flex block attention via `flex_block_attn_op`, using square 256×256 blocks (BLOCK_M = BLOCK_N = 256) rather than aiter_sparge's 256×128, since flex's rectangular-block path is slower/broken. It calls the shared sparge setup with `pad_block_divisible=True` so the query/key sequences are padded up to align = `lcm(block_m, block_n)` (computed inline via the Euclidean gcd to avoid a torch.compile graph break) to satisfy the flex kernel's divisibility requirements, with fully-padded key blocks zeroed out by mask_padded_kv_blocks and the padding stripped back off in `restore_sparge_output`; the new `pad_block_divisible `flag defaults to False so the AITER backends are unaffected. `runtime_state.py` wires up recognition/availability of the new backend, and `base_model.py` gates all sparge backends (flex included) to Wan only with cross-attention disallowed, so `text_len `is effectively zero in the supported config, the net result being a flex-based sparse backend that mirrors the sparge masking.

### DIFF
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -975,7 +975,7 @@ def _aiter_sparge_v2_attn_call(query, key, value, dropout_p, is_causal, attentio
 
 @register_attention_function(AttentionBackendType.FLEX_BLOCK_SPARGE)
 def _flex_block_sparge_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=None):
-    config = {"BLOCK_M": 128, "BLOCK_N": 128}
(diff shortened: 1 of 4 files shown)


avjves (approved):
LGTM!

PR #722, opened 2026-06-15
[Model] Add Cosmos3-Super/Nano support

nholmber:
## Summary

Add parallel inference support for [nvidia/Cosmos3-Super](https://huggingface.co/nvidia/Cosmos3-Super) (64B) and [nvidia/Cosmos3-Nano](https://huggingface.co/nvidia/Cosmos3-Nano) (16B) omnimodal world models. Supports both text-to-video (T2V) and image-to-video (I2V) generation.

Cosmos3 is architecturally distinct from existing xDiT models — it uses a Mixture-of-Transformers (MoT) with dual towers (autoregressive reasoner + diffusion generator) sharing joint attention within each layer. Text token IDs and vision latent tokens are packed into a single joint sequence with no separate text encoder.

### Features

- **USP** (Ulysses sequence parallelism) on the DM attention pathway with gen_seq chunking across ranks
- **CFG parallelism** with velocity-level gather in the denoising loop (cond/uncond on separate GPUs)
- **FSDP2** weight sharding across `transformer.layers`
- **FP8 GEMM** quantization via AITER blockscale
- **Mixed FP4/FP8 GEMM** precision (first/last 10 layers kept at FP8)
- **Parallel VAE** encode/decode via DistVAE (same `AutoencoderKLWan` as Wan models)
- **torch.compile** with inductor compute/comm overlap
- **I2V** support via `--input_images`
- **JSON prompt** file support (Cosmos3 structured caption format for best quality)

### Scope

This PR covers **video generation only** (T2V and I2V). Cosmos3 also supports audio, action/robotics, and text generation, which are left out of this PR.

### Attention backend note

Only full-precision attention backends are supported (`AITER`, `SDPA`, `FLASH`, `FLASH_3`). Quantized backends (`AITER_FP8`, `AITER_SAGE`, `AITER_SAGE_V2`, `AITER_MLA`, `FLASH_3_FP8`, `FLASH_4_FP4`, etc.) produce >50% relative error per attention layer on Cosmos3 due to extreme K/V dynamic range mismatch in the MoT dual-pathway attention (K absmax ~107 vs V absmax ~1.7). The model validates the backend at init and raises a clear error for unsupported backends. Quantized attention support to be added separately pending per-head or per-channel quantization.

## Performance

I2V, 720p, 121 frames (5s @ 24fps), 35 steps, 8x MI355X:

| Config | Total | Per-step | Speedup |
|--------|-------|----------|---------|
| BF16, U=8 | 30.9s | 0.88s | 1.0x |
| BF16, U=4 + CFG=2 | 24.0s | 0.69s | 1.3x |
| BF16, FSDP=8 + U=8 | 37.3s | 1.07s | 0.8x |
| BF16, FSDP=8 + U=4 + CFG=2 | 26.3s | 0.75s | 1.2x |
| FP8 GEMMs + U=4 + CFG=2 | 19.5s | 0.56s | 1.6x |

CFG parallel halves denoising work and U=4 gives fatter attention kernels (16 Q heads/GPU) with less all-to-all overhead than U=8. FP8 GEMMs add ~19% on top.

## Samples

Generated with `--ulysses_degree 4 --use_cfg_parallel --use_fp8_gemms`, 720p, 121 frames (5s @ 24fps), 35 steps, 8x MI355X:

**T2V** (prompt: `assets/example_t2v_prompt.json`):

https://github.com/user-attachments/assets/a6387abe-5c13-4070-8e7e-223e2ebdae33

**I2V** (prompt: `assets/example_i2v_prompt.json`, input: `assets/example_i2v_input.jpg`):

https://github.com/user-attachments/assets/b8ca673e-7154-44f5-a376-7f13fb2064e3

## Usage

The examples below use prompt/negative prompt JSON files bundled with the model (downloaded automatically with `huggingface-cli download nvidia/Cosmos3-Super`). Cosmos3 produces best quality with [JSON-upsampled prompts](https://huggingface.co/nvidia/Cosmos3-Super#prompt-upsampling); plain text prompts also work but with reduced quality.

```bash
# Download model assets (includes example prompts and input images)
huggingface-cli download nvidia/Cosmos3-Super

# Text-to-Video (8 GPUs: U=4 x CFG=2)
xdit --model nvidia/Cosmos3-Super \
  --ulysses_degree 4 --use_cfg_parallel \
  --use_torch_compile --use_parallel_vae --use_fp8_gemms \
  --prompt assets/example_t2v_prompt.json \
  --negative_prompt assets/negative_prompt.json \
  --num_frames 121 --height 720 --width 1280

# Image-to-Video
xdit --model nvidia/Cosmos3-Super \
  --ulysses_degree 4 --use_cfg_parallel \
  --use_torch_compile --use_parallel_vae --use_fp8_gemms \
  --input_images assets/example_i2v_input.jpg \
  --prompt assets/example_i2v_prompt.json \
  --negative_prompt assets/negative_prompt.json \
  --num_frames 121 --height 720 --width 1280

# With FSDP (reduced VRAM per GPU)
xdit --model nvidia/Cosmos3-Super \
  --ulysses_degree 4 --use_cfg_parallel --fully_shard_degree 8 \
  --use_torch_compile --use_parallel_vae \
  --prompt assets/example_t2v_prompt.json \
  --negative_prompt assets/negative_prompt.json \
  --num_frames 121 --height 720 --width 1280
```

Requires `diffusers >= 0.39.0` (for `Cosmos3OmniPipeline` and `Cosmos3OmniTransformer`).

## Files

| File | Description |
|------|-------------|
| `xfuser/model_executor/models/runner_models/cosmos3.py` | Runner model registration, loading, I2V support, FSDP/FP8/FP4 config |
| `xfuser/model_executor/models/transformers/transformer_cosmos3.py` | Transformer wrapper with dual-pathway attention processor and gen_seq chunking for SP |
| `xfuser/model_executor/pipelines/pipeline_cosmos3_omni.py` | Pipeline wrapper with CFG parallelism in the denoising loop |

No changes to existing files — auto-discovered via the runner model registry.

### DIFF
--- a/xfuser/model_executor/models/runner_models/cosmos3.py
+++ b/xfuser/model_executor/models/runner_models/cosmos3.py
@@ -1,0 +1,244 @@
+import copy
+import json
+import torch
+from diffusers import UniPCMultistepScheduler
+from diffusers.pipelines.pipeline_utils import DiffusionPipeline
+
+from xfuser import xFuserArgs
+from xfuser.model_executor.models.transformers.transformer_cosmos3 import (
+    get_cosmos3_transformer_wrapper_class,
+)
+from xfuser.model_executor.pipelines.pipeline_cosmos3_omni import (
+    get_cosmos3_pipeline_class,
+)
+from xfuser.model_executor.models.runner_models.base_model import (
+    ModelSettings,
+    xFuserModel,
+    register_model,
+    ModelCapabilities,
+    DefaultInputValues,
+    DiffusionOutput,
+    _parse_attention_backend,
+)
+from xfuser.core.distributed.parallel_state import get_vae_parallel_group
+from xfuser.core.distributed.attention_backend import AttentionBackendType
+from xfuser.core.utils.runner_utils import log, resize_and_crop_image
+
+# Only full-precision attention backends produce correct results on Cosmos3.
+# Quantized backends (FP8, MXFP4, MLA) cause >50% relative error per layer
+# due to extreme K/V dynamic range mismatch in the MoT attention.
+_COSMOS3_SUPPORTED_ATTN_BACKENDS = frozenset({
+    AttentionBackendType.AITER,
+    AttentionBackendType.SDPA,
+    AttentionBackendType.SDPA_MATH,
+    AttentionBackendType.SDPA_EFFICIENT,
+    AttentionBackendType.SDPA_FLASH,
+    AttentionBackendType.FLASH,
+    AttentionBackendType.FLASH_3,
+    AttentionBackendType.FLASH_4,
+})
+
+
+COSMOS3_FSDP_STRATEGY = {
+    "transformer": {
+        "wrap_attrs": ["layers"],
+        "dtype": torch.bfloat16,
+    },
+}
+
+
+def _setup_parallel_vae(vae, enable_parallel_encoder=True):
+    if enable_parallel_encoder:
+        try:
+            from distvae.modules.adapters.vae.encoder_adapters import WanEncoderAdapter
+            # scale_factor_spatial includes patch_size (e.g. 16 = 8x from encoder + 2x
+            # from patching). The encoder adapter needs just the encoder's own
+            # downsampling ratio, excluding the patching step.
+            vae_scale_factor = getattr(vae.config, 'scale_factor_spatial', 8)
+            patch_size = getattr(vae.config, 'patch_size', None)
+            if patch_size and patch_size > 1:
+                vae_scale_factor = vae_scale_factor // patch_size
+            patched_encoder = WanEncoderAdapter(
+                vae.encoder,
+                vae_group=get_vae_parallel_group().device_group,
+                vae_scale_factor=vae_scale_factor,
+            ).to(vae.device)
+            vae.encoder = patched_encoder
+            log("Parallel VAE encoder enabled.")
+        except ImportError:
+            log("DistVAE not available for encoder. Defaulting to single-rank.")
+        except Exception as e:
+            raise ValueError(f"Failed to patch VAE encoder: {e}")
+    try:
+        from distvae.modules.adapters.vae.decoder_adapters import WanDecoderAdapter
+        patched_decoder = WanDecoderAdapter(
+            vae.decoder, vae_group=get_vae_parallel_group().device_group
+        ).to(vae.device)
+        # Cosmos3 VAE has patch_size=2 (extra 2x spatial upsampling from
+        # unpatching). The decoder adapter's scale_factor defaults to 1 in
+        # its Patchify, which is correct for Wan (patch_size=None). For
+        # Cosmos3 we need to account for the extra factor so that the
+        # narrow in _forward crops to the right size.
+        patch_size = getattr(vae.config, 'patch_size', None)
+        if patch_size and patch_size > 1:
+            patched_decoder.patchify.scale_factor = patch_size
+        vae.decoder = patched_decoder
+        log("Parallel VAE decoder enabled.")
+    except ImportError:
+        log("DistVAE not available for decoder. Defaulting to single-rank.")
+    except Exception as e:
+        raise ValueError(f"Failed to patch VAE decoder: {e}")
+
+
+@register_model("nvidia/Cosmos3-Super")
+@register_model("Cosmos3-Super")
+class xFuserCosmos3SuperModel(xFuserModel):
+
+    capabilities = ModelCapabilities(
+        ulysses_degree=True,
+        ring_degree=True,
+        fully_shard_degree=True,
+        use_cfg_parallel=True,
+        use_parallel_vae=True,
+        use_parallel_vae_encoder=True,
+        use_fp8_gemms=True,
+        use_fp4_gemms=True,
+        enable_slicing=True,
+        enable_tiling=True,
+    )
+    default_input_values = DefaultInputValues(
+        height=720,
+        width=1280,
+        num_frames=189,
+        num_inference_steps=35,
+        guidance_scale=6.0,
+        flow_shift=10.0,
+    )
+    settings = ModelSettings(
+        model_name="nvidia/Cosmos3-Super",
+        output_name="cosmos3_super",
+        model_output_type="video",
+        fps=24,
+        mod_value=16,
+        fp8_gemm_module_list=["transformer.layers"],
+        fp4_gemm_module_list=["transformer.layers"],
+        fp8_precision_overrides=tuple(
+            f"{i}." for i in list(range(10)) + list(range(54, 64))
+        ),
+        fsdp_strategy=COSMOS3_FSDP_STRATEGY,
+    )
+
+    def _validate_config(self, config) -> None:
+        super()._validate_config(config)
+        backend = _parse_attention_backend(config.attention_backend, "attention backend")
+        if backend is not None and backend not in _COSMOS3_SUPPORTED_ATTN_BACKENDS:
+            supported = ", ".join(sorted(b.name for b in _COSMOS3_SUPPORTED_ATTN_BACKENDS))
+            raise ValueError(
+                f"Cosmos3 does not support --attention_backend {backend.name}. "
+                f"Quantized attention backends produce >50% relative error on this "
+                f"model due to extreme K/V dynamic range mismatch in the MoT attention. "
+                f"Supported backends: {supported}"
+            )
+
+    def _load_model(self) -> DiffusionPipeline:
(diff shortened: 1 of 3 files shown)


xfuser/model_executor/models/runner_models/cosmos3.py
```diff
@@ -0,0 +1,228 @@
...
+
+        kwargs = dict(
+            prompt=prompt,
+            negative_prompt=negative_prompt,
+            height=input_args["height"],
+            width=input_args["width"],
+            num_frames=input_args["num_frames"],
+            num_inference_steps=input_args["num_inference_steps"],
+            guidance_scale=input_args["guidance_scale"],
+            fps=float(input_args.get("fps", 24)),
```
avjves:
```suggestion
            fps=self.settings.fps,
```

FPS is not something that can be currently given from CLI nor is the saving reading it from input_args.

xfuser/model_executor/models/runner_models/cosmos3.py
```diff
@@ -0,0 +1,228 @@
...
+# Quantized backends (FP8, MXFP4, MLA) cause >50% relative error per layer
+# due to extreme K/V dynamic range mismatch in the MoT attention.
+_COSMOS3_SUPPORTED_ATTN_BACKENDS = frozenset({
+    AttentionBackendType.AITER,
+    AttentionBackendType.SDPA,
+    AttentionBackendType.SDPA_MATH,
+    AttentionBackendType.SDPA_EFFICIENT,
+    AttentionBackendType.SDPA_FLASH,
+    AttentionBackendType.FLASH,
+    AttentionBackendType.FLASH_3,
```
avjves, line 37: Should FLASH_4 also be here?

xfuser/model_executor/models/transformers/transformer_cosmos3.py
```diff
@@ -0,0 +1,267 @@
...
+                    action_noisy_frame_indexes=action_noisy_frame_indexes,
+                    action_domain_ids=action_domain_ids,
+                )
+
+            # SP path: with Ulysses parallelism, each rank has the full
+            # sequence but USP splits heads via all-to-all inside the attention.
+            # No sequence splitting needed at the transformer wrapper level —
+            # just delegate to parent forward. The xFuserCosmos3AttnProcessor
+            # handles the Ulysses all-to-all internally via USP().
+            return Cosmos3OmniTransformer.forward(
```
avjves: I'm a bit perplexed at how this works - why would there be no need for sequence splitting? There's no sequence splitting in the AttnProcessor either. If we don't shard before we call USP(), we'll end up with P x full sequence, right? So we are doing more work, not less.

A quick test on MI300X, 67s/iter with U=4, 140s/iter with U=8.

avjves (commented):
Thanks for the contribution! 🙏 

Looks good, but I think that SP implementation does not work as expected currently. Sharding should be done at some point prior to calling USP - left a comment about it.

xfuser/model_executor/models/runner_models/cosmos3.py
```diff
@@ -0,0 +1,228 @@
...
+@register_model("Cosmos3-Nano")
+class xFuserCosmos3NanoModel(xFuserCosmos3SuperModel):
+
+    settings = ModelSettings(
+        model_name="nvidia/Cosmos3-Nano",
+        output_name="cosmos3_nano",
+        model_output_type="video",
+        fps=24,
+        mod_value=16,
+    )
```
avjves, line 244: The settings object overwrites the inherited settings, so this does not have the GEMM and FSDP settings from it. I'm assuming these should be here then as well? Otherwise the capabilities should also be changed.

xfuser/model_executor/models/runner_models/cosmos3.py
```diff
@@ -0,0 +1,228 @@
...
+    },
+}
+
+
+def _setup_parallel_vae(vae, enable_parallel_encoder=True):
+    if enable_parallel_encoder:
+        try:
+            from distvae.modules.adapters.vae.encoder_adapters import WanEncoderAdapter
+            vae_scale_factor = getattr(vae.config, 'scaling_factor', 8)
+            if hasattr(vae.config, 'scale_factor_spatial'):
```
avjves: If this is the same AutoEncoderKLWan, should this check be the same as in Wan? i.e. `vae_scale_factor_spatial`, not `scale_factor_spatial`

xfuser/model_executor/models/transformers/transformer_cosmos3.py
```diff
@@ -0,0 +1,267 @@
...
+
+        def _install_xfuser_processors(self):
+            for layer in self.layers:
+                layer.self_attn.processor = xFuserCosmos3AttnProcessor()
+
+        def _patch_time_embedder_for_fsdp(self):
+            """Wrap time_embedder to auto-cast inputs, fixing FSDP dtype mismatch."""
+            original_forward = self.time_embedder.forward
+            def _cast_forward(sample, *args, **kwargs):
+                param_dtype = next(self.time_embedder.parameters()).dtype
```
avjves: Would this cause a graph break?

xfuser/model_executor/models/transformers/transformer_cosmos3.py
nholmber: Yeah you're right. The previous implementation delegated to the parent forward without chunking gen_seq, so MLPs/norms ran on the full sequence on every rank. Fixed: gen_seq is now chunked across SP ranks before the layer loop (like Wan), so MLPs/norms process 1/P of the sequence. USP handles head redistribution inside attention via all-to-all. und_seq stays replicated (short, AR pathway).

Updated I2V 720p/121f/35step results on 8x MI355X:

| Config | Before | After |
|--------|--------|-------|
| BF16, U=8 | 335.4s | 30.9s |
| BF16, U=4 + CFG=2 | 104.8s | 24.0s |
| FP8 GEMMs + U=4 + CFG=2 | 94.9s | 19.5s |

xfuser/model_executor/models/runner_models/cosmos3.py
nholmber: Added `FLASH_4` to the allowlist.

xfuser/model_executor/models/runner_models/cosmos3.py
nholmber: Fixed. Using `self.settings.fps` now.

xfuser/model_executor/models/runner_models/cosmos3.py
nholmber: Fixed.  `xFuserCosmos3NanoModel.settings` now includes `fp8_gemm_module_list`, `fp4_gemm_module_list`, and `fsdp_strategy`.

xfuser/model_executor/models/transformers/transformer_cosmos3.py
nholmber: Fixed. Here's what a trace looks like with FSDP U4 CFG2
image

xfuser/model_executor/models/runner_models/cosmos3.py
nholmber: Investigated. Neither Wan nor Cosmos3 `AutoencoderKLWan` configs have `vae_scale_factor_spatial`. Both have `scale_factor_spatial` (8 for Wan, 16 for Cosmos3). Wan's existing check for `vae_scale_factor_spatial` only works because the default of 8 happens to be correct.

However, parallel VAE is now **disabled** for Cosmos3. Cosmos3's VAE uses `WanResidualUpBlock` (not `WanUpBlock`) and `patch_size=2`, which distvae's `WanDecoderAdapter` doesn't handle correctly — it produces half-height output (240px instead of 480px). Needs a distvae update to support the residual up blocks and patch_size>1. Will address separately

nholmber:
Thanks for the review. Updated the PR based on feedback

xfuser/model_executor/models/runner_models/cosmos3.py
nholmber: Actually, parallel VAE is now working. The root cause of the earlier half-height output was that Cosmos3's `scale_factor_spatial=16` includes both the encoder's 8x downsampling AND the 2x from `patch_size=2`. We were passing 16 to `WanEncoderAdapter(vae_scale_factor=...)` but it should be 8 (encoder-only ratio). The encoder adapter uses this as the `downsampling_factor` in its narrow, so passing 16 discarded half the encoder output, producing wrong latents. Fix: divide by `patch_size` before passing to the adapter. No changes to distvae needed.

avjves (approved):
Thanks for the fixes, looks good now!

PR #726, opened 2026-06-22
RDNA4/gfx1201 General compile optimizations

pds-amd:
- When using FSDP, do explicit shard-aligned compilation. This reduces some awkward graph breaks when dynamo tries to trace across block boundaries.
- Some refactoring and centralization of the model compile logic for consistency and maintainability.
- `reduce-overhead` is significantly slower on gfx1201 in my testing - this adds a simple check to switch to `default` where relevant.

### DIFF
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -384,6 +384,8 @@ class xFuserModel(abc.ABC):
 
 
     def _get_compile_mode(self) -> str:
+        # Overrides should return "default" when PACKAGES_CHECKER._on_rdna4():
+        # CUDA graphs are slow on RDNA4.
         return "default"  # TODO: Configurable
 
     def _get_compile_dynamic(self) -> Optional[bool]:
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -383,14 +383,51 @@ class xFuserModel(abc.ABC):
                 raise ValueError(f"Model {self.settings.model_name} does not support distilled_transformer_path or distilled_transformer_2_path params.")
 
 
+    def _get_compile_mode(self) -> str:
+        return "default"  # TODO: Configurable
+
+    def _get_compile_dynamic(self) -> Optional[bool]:
+        return None  # torch default (auto)
+
+    def _get_compiled_pipe_components(self) -> List[str]:
+        return ["transformer"]
+
+    def _get_compile_warmup_steps(self, input_args: dict) -> Optional[int]:
+        return 2  # None = skip step reduction, run full warmup cycle
 
     def _compile_model(self, input_args: dict) -> None:
-        """ Compile the model using torch.compile."""
+        """Compile pipe components with torch.compile.
+
+        When FSDP is active (fully_shard_degree > 1), compiles each component's
+        FSDP-wrapped block lists individually (read from fsdp_strategy wrap_attrs)
+        to avoid dynamo tracing through FSDP2 forward_pre_hooks and fragmenting
+        the graph at every block boundary.
+        """
         torch._inductor.config.reorder_for_compute_comm_overlap = True
-        self.pipe.transformer = torch.compile(self.pipe.transformer, mode="default") # TODO: Configurable
-        # two steps to warmup the torch compiler
-        input_args["num_inference_steps"] = 2  # Reduce steps for warmup # TODO: make this more generic
-        self._run_timed_pipe(input_args)
+        mode = self._get_compile_mode()
+        dynamic = self._get_compile_dynamic()
+        for component_name in self._get_compiled_pipe_components():
+            component = getattr(self.pipe, component_name, None)
+            if component is None:
+                continue
+            if self.config.fully_shard_degree > 1:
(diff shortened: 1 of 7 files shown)


xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -383,14 +383,51 @@ def _validate_config(self, config: xFuserArgs) -> None:
...
+                continue
+            if self.config.fully_shard_degree > 1:
+                wrap_attrs = self.settings.fsdp_strategy.get(component_name, {}).get("wrap_attrs", [])
+                compiled_any = False
+                for attr in wrap_attrs:
+                    block_list = getattr(component, attr, None)
+                    if block_list is not None:
+                        for i in range(len(block_list)):
+                            block_list[i] = torch.compile(block_list[i], mode=mode, dynamic=dynamic)
+                        compiled_any = True
```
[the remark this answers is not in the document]
pds-amd: Good catch. Updated.

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -293,22 +291,15 @@ def _post_load_and_state_initialization(self, input_args: dict) -> None:
...
-        self._run_timed_pipe(input_args)
+    def _get_compile_mode(self) -> str:
+        # CUDA graphs incompatible with FBCache cross-step caching,
+        # and cause pathological re-captures on RDNA4.
+        if self.config.use_fbcache or PACKAGES_CHECKER._on_rdna4():
+            return "default"
+        return "reduce-overhead"
+
+    def _get_compile_dynamic(self) -> Optional[bool]:
+        return False
```
avjves, line 302: Why is this explicitly set to False for just this model? Should we default to False always - IIRC we don't want to dynamically compile?

xfuser/model_executor/models/runner_models/ltx.py
```diff
@@ -180,7 +181,8 @@ def _run_pipe(self, input_args: dict) -> DiffusionOutput:
 
     def _compile_model(self, input_args: dict) -> None:
         torch._inductor.config.reorder_for_compute_comm_overlap = True
-        self.pipe.transformer.compile_repeated_blocks(mode="reduce-overhead")
+        compile_mode = "default" if PACKAGES_CHECKER._on_rdna4() else "reduce-overhead"
```
avjves: nit: To be more uniform with other models, maybe call the `_get_compile_mode()`here as well and use that as the mode.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -383,14 +383,54 @@ def _validate_config(self, config: xFuserArgs) -> None:
                 raise ValueError(f"Model {self.settings.model_name} does not support distilled_transformer_path or distilled_transformer_2_path params.")
 
 
+    def _get_compile_mode(self) -> str:
+        return "default"  # TODO: Configurable
```
avjves, line 389: So default is not _always_ faster on rdna? I was just thinking if we could do the checking here.

avjves (approved):
Looks good! That compile change is a nice addition in general! Less copying :)

I left a few comments - can you take a look at them?

xfuser/model_executor/models/runner_models/ltx.py
pds-amd: Fixed! :)

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: It is always faster in my testing so far for this GPU. The main reason why I didn't put it here, is only ltx and flux use something other than default. All the other runner models use default.

I've added a comment - would that be sufficient?

xfuser/model_executor/models/runner_models/flux.py
pds-amd: I didn't want to change the existing behaviour.

Right now this is only set to False for flux, all the other models follow the default which is `None`

https://github.com/xdit-project/xDiT/blob/fe9e7e7e47d3496c36911e324b648c12749aa475/xfuser/model_executor/models/runner_models/flux.py#L308

https://github.com/xdit-project/xDiT/blob/fe9e7e7e47d3496c36911e324b648c12749aa475/xfuser/model_executor/models/runner_models/flux.py#L404

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Or would you prefer we just make it more explicit here like:

```python
  def _get_compile_mode(self) -> str:                                                                                                                                                                                                                                                                                                                               
      if PACKAGES_CHECKER._on_rdna4():                                                                                                                                                                                                                                                                                                                              
          return "default"                                                                                                                                                                                                                                                                                                                                          
      return "default"  # override to return a different mode    
```

xfuser/model_executor/models/runner_models/flux.py
avjves: Ah, makes sense :)
Seems like we added dynamic=False there for the FBCache. This is good!

xfuser/model_executor/models/runner_models/base_model.py
avjves: Ah, yeah - it's good now. Just unfortunate that we have to replicate that rdna check a lot in the models, but I suppose there's nothing to do about it.

PR #727, opened 2026-06-22
Add Head Balancing for AITER Sparge Backends

ksikiric:
This pull request introduces a new feature for block-sparse attention: per-rank head balancing for Ulysses parallelism, controlled by a new CLI/config flag. When enabled, attention heads are permuted so that each Ulysses rank receives a balanced subset of heads based on per-head computational cost, improving load balancing for Sparge attention backends. The implementation includes new logic for tracking, computing, and applying these permutations, as well as handling the necessary state and communication between ranks. The changes are backward-compatible and have no effect unless the new flag is set.

Key changes by theme:

**Feature: Ulysses Block-Sparse Head Balancing**
* Added a new config/CLI flag `use_spargeattn_head_balance` to enable per-rank head balancing for Sparge attention backends in distributed Ulysses setups.
* Implemented the `head_balance` module with logic for enabling/disabling the feature, computing balanced head permutations, and scattering per-head costs back to the global order.
* Integrated head balancing into the USP attention path: heads are permuted before all-to-all communication, per-head costs are gathered and exchanged, and the permutation is updated each step.
* Registered a per-layer, non-persistent buffer to track the current head permutation, ensuring compatibility with torch.compile and avoiding state_dict pollution.
* Ensured the head balancing feature is enabled/disabled before model compilation for correct torch.compile tracing.

**Integration and API Changes**
* Updated the attention and transformer code to accept and propagate the new `head_balance_layer` argument, ensuring the balancing logic is only applied where appropriate. 

**Backend Support**
* Modified the Sparge attention backend to publish per-head costs to the balancing logic via a new "cost sink" in `attention_kwargs` when balancing is enabled.

### DIFF
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -18,6 +18,7 @@ from xfuser.core.sparge_attention.sparge import (
     restore_sparge_output,
     mask_padded_kv_blocks,
 )
+from xfuser.core.sparge_attention import head_balance
 
 ATTENTION_FUNCTION_REGISTRY = {}
 
@@ -969,6 +970,14 @@ def _build_sparge_block_mask(query, key, value, is_causal, attention_kwargs, con
     )
     block_mask = mask_padded_kv_blocks(block_mask, state, config["BLOCK_N"])
     num_heads = q.shape[1]
+    # Per-head selected-block cost for the Ulysses head-balancer, written into
+    # the scratch "cost sink" tensor that USP passes down via attention_kwargs
+    # (only present when balancing is active). This keeps the head cost OUT of
+    # the (output, softmax_lse) return contract.
+    if head_balance.ENABLED:
+        cost_sink = (attention_kwargs or {}).get(head_balance.COST_SINK_KEY)
+        if cost_sink is not None:
+            cost_sink.copy_(block_mask.to(torch.float32).sum(dim=(0, 2, 3)))
     return q, k, v, state, block_mask, num_heads
 
 @register_attention_function(AttentionBackendType.AITER_SPARGE)
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -18,7 +18,7 @@ from xfuser.core.sparge_attention.sparge import (
     restore_sparge_output,
     mask_padded_kv_blocks,
 )
-from xfuser.core.sparge_attention import head_balance
+from xfuser.core.sparge_attention.head_balance import COST_SINK_KEY
 
 ATTENTION_FUNCTION_REGISTRY = {}
 
@@ -973,7 +973,7 @@ def _build_sparge_block_mask(query, key, value, is_causal, attention_kwargs, con
     # Per-head selected-block cost for the Ulysses head-balancer. USP injects a
     # scratch "cost sink" tensor into attention_kwargs only when balancing is
     # active.
-    cost_sink = (attention_kwargs or {}).get(head_balance.COST_SINK_KEY)
+    cost_sink = (attention_kwargs or {}).get(COST_SINK_KEY)
     if cost_sink is not None:
         cost_sink.copy_(block_mask.to(torch.float32).sum(dim=(0, 2, 3)))
     return q, k, v, state, block_mask, num_heads
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -970,14 +970,12 @@ def _build_sparge_block_mask(query, key, value, is_causal, attention_kwargs, con
     )
     block_mask = mask_padded_kv_blocks(block_mask, state, config["BLOCK_N"])
     num_heads = q.shape[1]
-    # Per-head selected-block cost for the Ulysses head-balancer, written into
-    # the scratch "cost sink" tensor that USP passes down via attention_kwargs
-    # (only present when balancing is active). This keeps the head cost OUT of
-    # the (output, softmax_lse) return contract.
-    if head_balance.ENABLED:
-        cost_sink = (attention_kwargs or {}).get(head_balance.COST_SINK_KEY)
-        if cost_sink is not None:
-            cost_sink.copy_(block_mask.to(torch.float32).sum(dim=(0, 2, 3)))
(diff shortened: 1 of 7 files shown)


xfuser/core/distributed/attention_backend.py
```diff
@@ -1017,6 +970,14 @@ def _build_sparge_block_mask(query, key, value, is_causal, attention_kwargs, con
     )
     block_mask = mask_padded_kv_blocks(block_mask, state, config["BLOCK_N"])
     num_heads = q.shape[1]
+    # Per-head selected-block cost for the Ulysses head-balancer, written into
+    # the scratch "cost sink" tensor that USP passes down via attention_kwargs
+    # (only present when balancing is active). This keeps the head cost OUT of
+    # the (output, softmax_lse) return contract.
+    if head_balance.ENABLED:
```
avjves: This global `ENABLED` flag is a bit weird. Would just checking if the `COST_SINK_KEY` is present be enough? Or enable it in runtime state and check from there. If it's enabled prior to compile, it should not be a problem, right? It's used in the usp to check the attention backend already.

xfuser/model_executor/layers/usp.py
```diff
@@ -241,17 +255,63 @@ def USP(
...
+    of heads (block-sparse load balancing); the permutation is inverted on the
+    output. No-op for non-sparse backends (no cost is published) and for ring/
+    joint paths.
     """
     if combine_qkv_a2a is None:
         combine_qkv_a2a = False
 
     attention_function = _get_attention_function(backend=backend)
 
+    # ---- Ulysses block-sparse head load balancing ----
```
avjves: This USP function is starting to be pretty complex as well 🤔 
But to ease the readabilty, can you at at least put the HB stuff into functions?

so it's something like:
```
if head_balance_layer and runtime_state.balance_ulysses_heads:
    head_balance_active = _balance_ulysses_heads(...)
...
if X:
   _revert_ulysses_head_balance(...)
```

avjves: These functions should probably live in the sparge_attention.py and be imported like `head_balance` is now

xfuser/model_executor/layers/usp.py
```diff
@@ -303,7 +363,7 @@ def USP(
                                         dropout_p=dropout_p,
                                         is_causal=is_causal,
                                         joint_attn_kwargs=joint_attn_kwargs,
-                                        attention_kwargs=attention_kwargs)
+                                        attention_kwargs=hb_attention_kwargs)
```
avjves: This just breaks non-sparge support, no? 
Head balancing args should just be inside attention_kwargs, not create a new variable.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -273,6 +256,11 @@ def initialize(self, input_args: dict) -> None:
 
     def _enable_options(self) -> None:
         """ Enable model options based on config"""
+        # Set the Sparge head-balancing gate before any torch.compile (which
+        # happens in _compile_model, called after this). Setting it here keeps
+        # head_balance.ENABLED a trace-time constant for dynamo.
+        head_balance.set_enabled(getattr(self.config, "use_spargeattn_head_balance", False))
```
avjves: Would be cleaner to just use an if clause + logging like VAE slicing below :)

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -204,6 +205,15 @@ def __init__(
         for block in self.blocks:
             block.attn1.processor = xFuserWanAttnProcessor(attention_kwargs=self.attention_kwargs)
             block.attn2.processor = xFuserWanAttnProcessor(use_ulysses_parallel_attention=False, is_cross_attention=True)
+            # Per-layer head permutation buffer for the Ulysses block-sparse head
+            # balancer (read/updated in-place inside USP; identity = no balancing).
+            # Registered pre-compile and non-persistent so it stays out of the
+            # state_dict and is captured as a graph input by torch.compile.
+            block.attn1.register_buffer(
```
avjves, line 212: Will this affect non-sparge runs? 😄 I gotta say I'm not 100% sure what it does

xfuser/model_executor/layers/usp.py
ksikiric: if hb_active == False, then hb_attention_kwargs will just be attention_kwargs. L296 in this file

xfuser/model_executor/models/transformers/transformer_wan.py
ksikiric: It is just a place to store the head permutations, so for the other backends, nothing will be placed there, and that particular buffer will never be used

xfuser/model_executor/layers/usp.py
avjves: Ahh, true, I missed that. But maybe it's still better to not name the generic path to get "hb_attention_kwargs", would indicate that they are all related to head balancing :)

xfuser/model_executor/layers/usp.py
```diff
@@ -241,17 +255,55 @@ def USP(
...
+    output. No-op for non-sparse backends (no cost is published) and for ring/
+    joint paths.
     """
     if combine_qkv_a2a is None:
         combine_qkv_a2a = False
 
     attention_function = _get_attention_function(backend=backend)
 
+    # ---- Ulysses block-sparse head load balancing ----
+    # Permute the heads before the input all-to-all so each rank receives a
```
avjves: Could all of these checks still be within that sparge attention file?
So it's just
```
query, key, value, hb_applied, attention_kwargs = apply_head_balance(...)
...
if hb_applied:
    out = revert_head_balance(out, attention_kwargs)
```

This head balancing is specific to spargeattn, so I'd prefer to not "pollute" the generic function too much. Keeping the balance params just in the attention_kwargs seems like valid to me. It obfuscates things a bit, but at the same time it keeps the generic function tidier.

avjves: nit: why does the attention_kwargs get renamed at all? 😄

ksikiric: Sorry, I rushed a bit much with this.. 😅

avjves (approved):
Looks good now, thanks :)

PR #734, opened 2026-07-09
fix: fix version comparisons using packaging.version.parse

Qin-sx:
Three version checks used string comparison, which is incorrect for multi-digit minor versions (e.g., `"0.100" >= "0.32"` returns `False` since `'1' < '3'` lexicographically).


jcaraban (approved):
Thanks!

PR #735, opened 2026-07-11
fix: fix pipefusion NCCL error

Qin-sx:
### Problem

PR https://github.com/xdit-project/xDiT/pull/731 added `device_id=envs.get_device(local_rank)` to `init_process_group()`. This breaks when running pipefusion + ulysses:

```
ft_c.all_to_all_single(x, ..., group=PROCESS_GROUP.ULYSSES_PG)
RuntimeError: NCCL Error 1: unhandled cuda error
```

### Cause

Passing `device_id=` to `init_process_group()` makes the default group use device-aware NCCL initialization. However, subgroups (pipefusion pipeline groups, ulysses/ring groups) are created via `new_group()` without passing `device_id`. yunchang's `set_seq_parallel_pg()` creates `ULYSSES_PG` and `RING_PG` without passing `device_id` — this cannot be fixed since yunchang is an external library. This initialization mode mismatch causes NCCL Error 1.

### Fix

Removed `device_id=` from `init_process_group()` and call `set_device(local_rank)` before it. This ensures the device is correctly set before any NCCL initialization, works for both the default group and all `new_group()` subgroups, without requiring changes to yunchang.

### DIFF
--- a/xfuser/core/distributed/parallel_state.py
+++ b/xfuser/core/distributed/parallel_state.py
@@ -160,6 +160,7 @@ def get_data_parallel_rank():
     """Return my rank for the data parallel group."""
     return get_dp_group().rank_in_group
 
+
 # FS (Fully Shard)
 def get_fs_group() -> GroupCoordinator:
     assert _FS is not None, "fully sharding group is not initialized"
@@ -253,6 +254,12 @@ def init_distributed_environment(
             local_rank = envs.LOCAL_RANK
         else:
             local_rank = rank
+    # Set the device BEFORE init_process_group so that NCCL communicators
+    # (both the default group and subsequent new_group() subgroups) bind
+    # to the correct device. Passing device_id= to init_process_group()
+    # instead causes NCCL to use device-side rendezvous, which breaks
+    # new_group() calls used by pipefusion and ulysses (NCCL Error 1).
+    set_device(local_rank)
     if not torch.distributed.is_initialized():
         assert distributed_init_method is not None, (
             "distributed_init_method must be provided when initializing "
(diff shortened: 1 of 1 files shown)


Qin-sx:
Hi @Arech8, could you have a look at whether this change affects the goals of PR [#731](https://github.com/xdit-project/xDiT/pull/731)? I don't have a multi-node environment to verify.

Arech8:
Thank @Qin-sx for sharing, that's unexpected... I'll check it! ~~Also not sure, why AMD/ROCm isn't included in the change? Could the `set_device()` be unconditional, since it looks like this might apply to all stacks?~~ EDIT: all right, looked up the `envs.get_device_name()`, hip is counted as cuda, so that's accounted for; Still not sure how to deal with `cpu` device and if it's relevant here.

I'll test on h100 rig and will find something with AMD to test it there too, though this would take some time...

Arech8:
Great, this works on 8xH100 rig, on SD3.5 model that was failing without device_id specification prior to #731.

Just to be extra sure, I'll find an AMD rig and try it there also.

Qin-sx:
> Thank @Qin-sx for sharing, that's unexpected... I'll check it! ~Also not sure, why AMD/ROCm isn't included in the change? Could the `set_device()` be unconditional, since it looks like this might apply to all stacks?~ EDIT: all right, looked up the `envs.get_device_name()`, hip is counted as cuda, so that's accounted for; Still not sure how to deal with `cpu` device and if it's relevant here.
> 
> I'll test on h100 rig and will find something with AMD to test it there too, though this would take some time...

Thanks for testing! Regarding the CPU device — I couldn't find a CPU launch path in xDiT. Also, set_device is a GPU-specific API, so skipping it for CPU is the correct behavior.

Arech8:
sorry for a delay, there are still some issues with setting up the test on AMD, working on resolution. Will post here once done...

Arech8:
all right, no issues on MI300x with SD3.5 with pytorch pre2.13 either here. Thanks!

cc @nsakkine , @jcaraban , - I think we should merge this PR

jcaraban (approved):
Thanks again @Qin-sx

PR #737, opened 2026-07-16
feat(compat): semantic diffusers-version gating + lazy pipeline imports

pds-amd:
A diffusers-version mismatch should degrade to "that one model is unavailable," never crash the package.

Several pieces work together:

- **Lazy imports.** The `from diffusers import <Pipeline>` lines move out of module scope and into each model's `_load_model()`, and runner modules no longer import their transformer wrappers at module scope either. A model's diffusers requirement is only pulled in when you actually load that model.
- **Graceful skip at registration.** The runner-package loader and the pipeline package skip a module whose diffusers symbols are missing, leaving its models unregistered rather than taking the whole import down. An `ImportError` that did not originate in diffusers is re-raised, so an xfuser bug still surfaces as an error instead of a misleading version warning. This replaces the old `config/diffusers.py` `has_valid_diffusers_version(...)` gate.
- **Per-model floor, used only for the message.** Each runner declares a `min_diffusers_version`, measured against real diffusers wheels. It never gates a load. Its one job is to turn a failed import into "requires diffusers>=0.37.0" rather than a stack trace. Where support has not shipped in any release the floor reads `DIFFUSERS_FROM_SOURCE`, and the message says to build diffusers from source.
- **One place to write a dependency floor.** `declared_floor()` in `xfuser/compat.py` reads a dependency's lower bound from xfuser's package metadata, which setuptools copies from setup.py's `install_requires`. When that metadata is unavailable the check is skipped; pip enforced the floor at install time anyway. No floor gets a hardcoded fallback.
- **Prerelease tolerance.** Version comparisons use PEP 440 release tuples, so `0.38.0.dev0` satisfies a `0.38.0` floor and a torch 2.6 nightly still takes the torch 2.6 code path.
- **FLUX.2 keeps working on pre-0.38 diffusers.** `Flux2Transformer2DModelOutput` only exists in 0.38; earlier releases return the generic `Transformer2DModelOutput` from the same method. The wrapper now follows whichever the installed version has, so FLUX.2 dev still gates at 0.36 and klein at 0.37, the releases their pipelines actually shipped in.
- **An exported pipeline comes with a wrapper for its backbone.** `from_pretrained` looks the backbone up in the wrapper registry, so a pipeline exported without that registration failed deep inside `from_pretrained` with no mention of diffusers. The FLUX.2 pipeline module imports its wrapper, and a test checks the pairing for every exported pipeline.

Tests in `tests/core/` cover this without weights or network access. The declared floors are audited against the installed diffusers, including a check that a `DIFFUSERS_FROM_SOURCE` marker has not gone stale now that the support has shipped. Runner modules are checked by AST for module-scope wrapper imports, and every exported pipeline is checked to have a registered wrapper for its backbone.


jcaraban:
❗AI assisted review❗ 👇 

### What's good
- Colocating the floor with each runner instead of a central `MINIMUM_DIFFUSERS_VERSIONS` dict is the right direction.
- The `e.name` re-raise guard works: I verified `from diffusers import TotallyFakePipeline` raises `ImportError` with `e.name == 'diffusers'`, so genuine xfuser bugs do propagate.
- Dropping the `>= 0.32` guard around the Flux `vae_scale_factor *= 2` in runtime_state.py is correct given the 0.33 install floor.

### Issues

**1. `min_diffusers_version` is unreachable for almost every model it was added to.**
The `from diffusers import <Pipeline>` lines moved into `_load_model()`, but the runner modules still eagerly import their transformer wrapper at module scope, and *those* import the version-specific diffusers symbols:

```
z_image.py:3  →  transformer_z_image.py:6  →  ZImageTransformer2DModel   (0.36+)
flux.py:4     →  transformer_flux.py:8     →  transformer_flux symbols
ltx.py:5      →  transformer_ltx2.py:3     →  transformer_ltx2 symbols
```

So on diffusers 0.35 the `z_image` module import still fails, the new `try/except` in __init__.py swallows it, and the model never registers. The user gets "unknown model", not "Z-Image requires diffusers>=0.36.0" — exactly the case the feature was written for. `krea2` is the only one that works, because it lazily imports its wrapper too. Either lazify the wrapper imports as well, or move the floor lookup to registration time so the friendly message can be emitted for a skipped module.

**2. Prerelease regression — this is the one I'd most want changed.**
The deleted `has_valid_diffusers_version` compared `.release` tuples *deliberately*, so `0.36.0.dev0` satisfied a `0.36.0` floor. The new check is a plain `Version` compare, and I verified `Version('0.36.0.dev0') < Version('0.36.0')` is `True` — so anyone on diffusers-from-source (the historically recommended install for xDiT; this box has `0.38.0.dev0`) gets locked out of a model they can actually run.

Same class of bug, worse failure mode, in `usp.py`: `parse(torch.__version__).release >= parse("2.6.0").release` became `Version(torch.__version__) >= Version("2.6.0")`. Verified `Version('2.6.0.dev20250101+rocm6.3') >= Version('2.6.0')` is `False` — a torch 2.6 nightly now silently skips `_cp_options.enable_load_balance = False`. That's a silent behavior change, not an error. Please keep the `.release` comparisons.

Worth noting the two mechanisms now disagree: I confirmed `'0.36.0.dev0' in SpecifierSet('>=0.33.0')` is `True`, so `check_packages` accepts dev builds while `_require_diffusers_version` rejects them.

**3. Missing floors.** Audited every runner class: `xFuserHunyuanvideo15SparseModel` lazily imports `HunyuanVideo15ImageToVideoPipeline` (0.36-only) with no floor, and `xFuserWan22T2VModel` / `xFuserWan22TI2VModel` have none while their Wan2.1 siblings do.

**4. `required_dist_specifier()` — complexity without the payoff.** It reads *installed* metadata; `requires('xfuser')` here returns `['diffusers>=0.33.0', 'flash-attn>=2.6.0; extra == "flash-attn"']`, which is a snapshot from install time. An editable install with stale egg-info reports the old floor. And because a source fallback is needed, the hardcoded `0.33.0` and `2.6.0` still exist — so the duplication it removes is replaced by duplication in two places plus a branch. I'd drop it. Minor: name matching should use `packaging.utils.canonicalize_name` (`flash-attn` vs `flash_attn`), and `PackageNotFoundError` is bound inside the `try` so the `except` clause would `NameError` if the metadata import itself failed.

**5. `registry.py`.** The outer `try` now also wraps the flux2 block, so an `ImportError` from the *xfuser* flux2 wrapper is silently swallowed — inconsistent with the `e.name` discipline applied in the two __init__.py files. Also, `FluxTransformer2DModel` has existed since diffusers 0.30, so with a 0.33 floor that gate is dead code.

**6. `self.settings = copy.deepcopy(self.__class__.settings)`.** The description calls this a latent-bug fix, but at that point in `__init__` there is no instance attribute, so `self.settings` already resolved to the class attribute and was already deep-copied. It's a semantic no-op — fine to keep for explicitness, but the claim should come out of the description.

jcaraban:
Only 1) and 2) seem more serious to me? but maybe we need a separate PR to harmonize the import checks all around?

xfuser/model_executor/models/transformers/__init__.py
```diff
@@ -21,20 +20,31 @@
     "xFuserSanaTransformer2DWrapper",
 ]
 
-# Gating some imports based on diffusers version, as they import part of diffusers
-if has_valid_diffusers_version("flux"):
```
avjves, line 25: I believe we could just remove these altogether - most transformers are not added here. These are here just so that one could import tne transformer directly, `from xfuser.model_executor.models.transformers import  xFuserKrea2Transformer2DWrapper`, but AFAIK we don't do that.

xfuser/model_executor/models/runner_models/flux.py
```diff
@@ -1,18 +1,7 @@
...
-    Flux2KleinPipeline,
-)
 from diffusers.pipelines.pipeline_utils import DiffusionPipeline
-from xfuser.model_executor.models.transformers.transformer_flux import (
-    xFuserFlux1Transformer2DWrapper,
-)
-from xfuser.model_executor.models.transformers.transformer_flux2 import (
-    xFuserFlux2Transformer2DWrapper,
-)
+from xfuser.model_executor.models.transformers.transformer_flux import xFuserFlux1Transformer2DWrapper
```
avjves: Shouldn't this be lazy imported as well? The Flux transformer is really old though, so maybe not a problem

avjves:
Left a couple of extra comments, but the jcaraban's AI review is spot on - 1) and 2) are critical to fix. Rest are either not true or less serious. I would address 4) and 5) :)

avjves:
In general I'm also not too fan of inline lazy importing, as they clutter the codebase and can lead into dependency issues, but maybe that's not too relevant nowadays. In Python 3.15 we'll get the `lazy` keyword which would fix all these, but unfortunately we are not there yet :) I think our current options are lazy importing within the functions or having a global list of model-specific diffusers reqs and I do believe the former is better.

xfuser/model_executor/models/transformers/__init__.py
pds-amd: I agree. They've been removed.

xfuser/model_executor/models/runner_models/flux.py
pds-amd: Moved, along with every other module-scope wrapper import in the runner modules.

pds-amd:
@jcaraban @avjves I have addressed the review comments, thanks!

I am also not a fan of lazy imports but it seems like the least bad choice for now and is a better UX than what's on main today.

1. Wrapper imports are lazy in every runner module now.
2. One `version_at_least()` comparing release tuples, used for diffusers, torch (usp.py and attention_processor.py), flash-attn and transformers. The two mechanisms you found disagreeing are the same function now.
3. Some have missing floors if they were available in the base diffusion version or before. Also bleeding edge models (that haven't landed in an official release version yet) will direct a user to installing diffusers from source.
4. `required_dist_specifier()` is gone, replaced by `declared_floor()` in `xfuser/compat.py`
5. Split into two sibling `try` blocks, each with the origin check, so an xfuser ImportError from either wrapper no longer disappears into a bare `pass`. On the dead code: the block also imports the xfuser wrapper, which needs `_get_qkv_projections`, `FluxAttention` and `FluxAttnProcessor` — none of which exist before 0.35.2 — so the guard does fire on 0.33 and 0.34.
6. Removed.

jcaraban (approved):

PR #738, opened 2026-07-16
feat(attention): gfx1201 flydsl fp8 flash-attention

pds-amd:
Adds an fp8 flash-attention backend (`AITER_FLYDSL_FP8`) for AMD RDNA4 (gfx1201) via flydsl, wired into the attention backend selector.

Also improves upon the existing bf16 `AITER_FLYDSL` attention backend.

Requires: https://github.com/ROCm/aiter/pull/4188 See further detail there on performance.


xfuser/core/distributed/attention_backend.py
```diff
@@ -380,17 +383,99 @@ def _build_aiter_mla_metadata(batch_size, q_seq_len, kv_seq_len, num_heads, num_
...
+            # flydsl_fp8_quant returns fp8 q/k/v + descales (real = fp8 * descale). Imported here,
+            # not at module load: fp8 quant ships in a newer aiter than the bf16 kernel, so a
+            # bf16-only build still registers the bf16 op above; runtime_state gates fp8 selection.
+            from aiter.ops.flydsl import flydsl_fp8_quant as flydsl_fp8_quant_aiter
+            qq, kk, vv, sq, sk, sv = flydsl_fp8_quant_aiter(query, key, value, rotation=True)
+            return flydsl_flash_attn_func_aiter(
+                qq, kk, vv, causal=is_causal,
+                q_descale=sq, k_descale=sk, v_descale=sv,
+                waves_per_eu=2, daz=True,
+            )
```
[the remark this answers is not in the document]
pds-amd: Top-level import would fail the whole block if flydsl_fp8_quant isn't present.  See existing comment.

xfuser/core/distributed/attention_backend.py
```diff
@@ -1158,21 +1244,37 @@ def _flex_block_sparge_attn_call(query, key, value, dropout_p, is_causal, attent
...
-    # when the seq_len padding ratio to align to 128 exceeds 0.5% (same check as the kernel).
-    if query.shape[2] != key.shape[2]:
+def _aiter_flydsl_dispatch(query, key, value, dropout_p, is_causal, attention_kwargs, attn_op):
+    # Layout here is [B, H, S, D]. Self-attn and non-causal cross-attn both hit the
+    # kernel: causal masks padded cols via col > q_row; non-causal masks them via
+    # seq_len_real (aligned) or the tail mask (unaligned); cross-attn (Phase B) loads
+    # K/V on their own length via seq_len_kv. The kernel is single-NUM_HEADS MHA, so
+    # only causal-cross (ambiguous alignment; never occurs in diffusion) and GQA / head
+    # or head_dim mismatches fall back to SDPA.
+    is_cross = query.shape[2] != key.shape[2]
```
jcaraban, line 1269: 🤖  The comment above says "GQA / head or head_dim mismatches fall back to SDPA", but the head-count and head-dim checks are nested under `is_cross`, so they only run for cross-attention. GQA **self**-attention (`S_q == S_k`, `H_q != H_k`) skips the fallback and goes straight into a kernel the comment itself describes as single-`NUM_HEADS` MHA.

The pre-#4188 kernel happens to catch this via its `q.shape == k.shape == v.shape` assertion, but a build that relaxes that check to support cross-attention may not. Suggest hoisting the head checks out of the conjunction:

```suggestion
    is_cross = query.shape[2] != key.shape[2]
    if (
        query.shape[1] != key.shape[1]
        or query.shape[3] != key.shape[3]
        or (is_cross and is_causal)
    ):
```

xfuser/core/distributed/runtime_state.py
jcaraban, line 316: 🤖  This probe passes on pre-#4188 AITER builds, where the kernel still rejects cross-attention and >0.5%-padded non-causal shapes — the two cases whose SDPA fallbacks this PR removes. Since `AITER_FLYDSL` is auto-selected on RDNA4, that turns into an unexplained `ValueError` mid-run for users who never opted in.

Suggest probing for the capability rather than just the symbol:

```python
elif attention_backend == AttentionBackendType.AITER_FLYDSL:
    try:
        from aiter.ops.flydsl import flydsl_flash_attn_func
    except ImportError:
        raise RuntimeError("AITER FlyDSL attention is not available, please update AITER") from None
    # Cross-attn / unaligned non-causal support landed with seq_len_kv; older builds raise.
    import inspect
    if "seq_len_kv" not in inspect.signature(flydsl_flash_attn_func).parameters:
        raise RuntimeError(
            "AITER FlyDSL attention requires a build with cross-attention support, please update AITER"
        ) from None
```

(Adjust the parameter name to whatever #4188 actually lands.) Alternatively, just keep the two fallbacks in `_aiter_flydsl_dispatch` — they cost one shape comparison and degrade instead of crashing.

jcaraban:
🤖 Nice split into two ops — mirroring the existing `AITER` / `AITER_FP8` pattern makes this easy to follow, and the wiring is complete (enum lookup means the CLI picks up `aiter_flydsl_fp8` for free, the low-precision warning and ring-degree guard lists are both updated, and there are no stale references to the old `xfuser::flydsl_attn_kernel` name).

Two things I'd like addressed before merge, both on the **shared bf16 path** rather than the new fp8 code.

**1. The SDPA fallbacks were removed from `AITER_FLYDSL`, but the availability probe wasn't tightened to match.**

`main` fell back to SDPA for (a) cross-attention and (b) non-causal self-attention with a >0.5% alignment-padding ratio. This PR drops both, on the basis that ROCm/aiter#4188 makes the kernel handle them. But `_check_attention_backend_available` still only probes that `flydsl_flash_attn_func` is *importable*, which is equally true on pre-#4188 builds.

I checked against an AITER that predates #4188 (`aiter/ops/flydsl/fmha_kernels.py`):

```python
def flydsl_flash_attn_func(q, k, v, causal=False, waves_per_eu=2, daz=True, stream=None)
    ...
    if not (q.shape == k.shape == v.shape):
        raise ValueError("flydsl_flash_attn_func is self-attention; q/k/v must share shape, ...")
    ...
    if not causal and n_pad > 0 and n_pad / seq_len_pad > _MAX_NONCAUSAL_PAD_RATIO:
        raise ValueError("... padding ratio ... exceeds 0.5% safety threshold ...")
```

So with that AITER installed, this PR converts both fallbacks into hard `ValueError`s. That's amplified by `AITER_FLYDSL` being the **auto-selected default on RDNA4** (`_select_attention_backend`: `if env_info["has_aiter"] and PACKAGES_CHECKER._on_rdna4(): backend = AttentionBackendType.AITER_FLYDSL`) — so this reaches users who never explicitly chose flydsl, and it presents as a crash rather than a "please update AITER" message.

Either keep the fallbacks (they're a single shape comparison, and can only fire in cases where the kernel would have raised anyway), or make the probe detect the new capability — see the inline suggestion on `runtime_state.py`.

Good news: `waves_per_eu` and `daz` already exist with those exact defaults in the old signature, so those added kwargs are safe on older builds.

**2. The GQA guard doesn't do what its comment says** — the head-count / head-dim checks are nested under `is_cross`, so GQA *self*-attention bypasses them. Inline below.

The rest is comment-level. One broader ask: the description links #743 for performance, but the README this PR touches warns that fp8 attention can introduce visual artifacts. Could you add some accuracy evidence for the fp8 path (cos-sim against bf16, or a reference image/video pair)? That would make this much easier to approve, especially given the crossover heuristic silently changes precision based on sequence length.

xfuser/core/distributed/runtime_state.py
pds-amd: Good catch. I now detect the capability once and keep the fallbacks only for builds that need them.

xfuser/core/distributed/attention_backend.py
pds-amd: Fixed.

pds-amd:
Thanks for the review!

On the performance questions

1. Are the cosine similarity numbers under the collapsed details section here enough? https://github.com/ROCm/aiter/pull/4188 Otherwise let me know if you want to see any specific experiments here.

2. Regarding visual artifacts - the only real issue I encountered was with Wan having some blocky noise (in the cat's fur for example), image models look identical to me when switching the attention backend.

### FLUX.2-dev BF16 Attention Example

<details>

```
    "seed": 42,
    "prompt": "Cat astronaut raising a cat flag on the surface of moon",
    "height": 1024,
    "width": 1024,
    "num_inference_steps": 50,
    "max_sequence_length": 512,
    "warmup_calls": 1,
    "num_iterations": 3,
    "guidance_scale": 4.0,
    "attention_backend": "aiter_flydsl",
    "ulysses_degree": 4,
    "fully_shard_degree": 2,
    "memory_efficient_sharding": true,
    "use_torch_compile": true,
    "use_fp8_gemms": true,
    "cache_method": "dbcache"
```

</details>

flux_2_dev_u4r1_tc_True_1024x1024_0

### FLUX.2-dev FP8 Attention Example

<details>

```
    "seed": 42,
    "prompt": "Cat astronaut raising a cat flag on the surface of moon",
    "height": 1024,
    "width": 1024,
    "num_inference_steps": 50,
    "max_sequence_length": 512,
    "warmup_calls": 1,
    "num_iterations": 3,
    "guidance_scale": 4.0,
    "attention_backend": "aiter_flydsl_fp8",
    "ulysses_degree": 4,
    "fully_shard_degree": 2,
    "memory_efficient_sharding": true,
    "use_torch_compile": true,
    "use_fp8_gemms": true,
    "cache_method": "dbcache"
```

</details>

flux_2_dev_u4r1_tc_True_1024x1024_0


### Wan2.2-TI2V-5B-Diffusers BF16 Attention Example

<details>

```
    "seed": 42,
    "prompt": "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds.",
    "height": 736,
    "width": 1280,
    "num_frames": 121,
    "num_inference_steps": 50,
    "warmup_calls": 2,
    "num_iterations": 5,
    "guidance_scale": 5.0,
    "flow_shift": 5,
    "task": "i2v",
    "input_images": "/app/data/wan_input.jpg",
    "attention_backend": "aiter_flydsl",
    "use_torch_compile": true,
    "use_fp8_gemms": true,
    "cache_method": "dbcache",
    "enable_tiling": true
```

</details>


https://github.com/user-attachments/assets/34a07b29-f77c-4fb5-8158-77613042b8eb



### Wan2.2-TI2V-5B-Diffusers FP8 Attention Example

<details>

```
    "seed": 42,
    "prompt": "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds.",
    "height": 736,
    "width": 1280,
    "num_frames": 121,
    "num_inference_steps": 50,
    "warmup_calls": 2,
    "num_iterations": 5,
    "guidance_scale": 5.0,
    "flow_shift": 5,
    "task": "i2v",
    "input_images": "/app/data/wan_input.jpg",
    "attention_backend": "aiter_flydsl_fp8",
    "use_torch_compile": true,
    "use_fp8_gemms": true,
    "cache_method": "dbcache",
    "enable_tiling": true
```

</details>

https://github.com/user-attachments/assets/206a95bc-977c-45aa-9570-04b68961c436

jcaraban (approved):

avjves (approved):

PR #739, opened 2026-07-16
gfx1201 VRAM + host memory management

pds-amd:
Systems with low VRAM often also have host system memory (RAM) restrictions as well.

On multi-GPU configuration, model load can currently have a transient spike of GPU_COUNT x MODEL_SIZE even if the model can fit just fine on the GPUs within the bounds of the provided parallelism parameters.

This adds several features that allow us to load large models even if system memory is tight.

Iteratively testing this at scale across a large benchmark sweep is very time consuming, so several interconnected optimizations have been bundled here.

1. Memory efficient meta load
    - Do load on rank0 and broadcast to the other GPUs block-wise
2. Streaming FP8 quantization on load
    - Do quantization as weights stream off disk so we never have to materialize the full model if quantization is enabled
3. Text Encoder Quantization
    - BF16 text encoders are still relatively huge for low VRAM card and we can quantize them to FP8 with no perceivable degradation
4.  Per-layer CPU eviction
    - Each quantized FP8 layer is immediately offloaded to host RAM, so weights don't accumulate on GPU during load either to keep VRAM available for large quantization steps.
5. `--enable_group_cpu_offload`
    - new CLI flag enabling group-granularity CPU offload for single GPU inference. Faster than sequential offload, slower than model offload.
6. Page-cache drop after shard reads
    - after rank0 reads each weight shard, it advises the kernel to evict the file from page cache. Prevents the OS from double-counting shard data against host RAM while the model is resident.

Minor fixes:
1. `_make_generator` CPU-offload fix
    - minor generator device placement hardening.
2. z_image FSDP wrap_attrs bug fix `model.language_model.layers` should be `layers`.


xfuser/model_executor/models/runner_models/meta_load.py
```diff
@@ -0,0 +1,596 @@
...
+        real submodules literally named 'module' are left intact.
+        """
+        from xfuser.model_executor.base_wrapper import xFuserBaseWrapper
+        cur, out = root, []
+        for seg in name.split("."):
+            if seg == "module" and isinstance(cur, xFuserBaseWrapper):
+                cur = getattr(cur, seg)
+                continue
+            out.append(seg)
+            cur = getattr(cur, seg)
```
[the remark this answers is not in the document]
pds-amd: `nn.Module.__getattr__` walks `_modules`, which stores children under their string keys ("0", "1", …). `getattr(module_list, "0")` returns the same object as `module_list[0]`. No fix needed.

xfuser/model_executor/models/runner_models/meta_load.py
```diff
@@ -0,0 +1,596 @@
...
+                    set_module_tensor_to_device(
+                        component, name, device,
+                        value=torch.empty(shape, dtype=dtype, device=device),
+                        dtype=dtype,
+                    )
+        for name in ordered:
+            # .contiguous() on the src: rank0's from_pretrained real tensors are trusted as-is and a
+            # non-contiguous buffer (strided/transposed view) would broadcast the wrong element bytes
+            # into peers. No-op on peers (freshly materialized -> already contiguous, filled in place).
+            world.broadcast(rgetattr(component, name).data.contiguous(), src=0)
```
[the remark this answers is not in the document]
pds-amd: `.contiguous()` on an already-contiguous tensor returns self (same storage pointer), so NCCL fills the original buffer in-place on peers. The comment in the code explains this explicitly. The suggestion would change nothing functionally. No fix needed.

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -80,6 +189,38 @@ def load_and_quantize_weights(
...
+
+        Stored as a plain __dict__ attribute (not a param/buffer): it must stay invisible to
+        named_parameters/named_buffers so it neither inflates the replicated-broadcast tensor count
+        (rank0 vs peers must agree) nor collides with the DiT quantizer's later `weight` handling.
+        0-element ⇒ no VRAM.
+        """
+        self._parameters.pop("weight", None)
+        self._buffers.pop("weight", None)
+        dev = self.weight_fp8.device if getattr(self, "weight_fp8", None) is not None else None
+        self.__dict__["weight"] = torch.empty(0, dtype=self._compute_dtype, device=dev)
```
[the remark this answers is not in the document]
pds-amd: The sentinel is only installed by `load_and_quantize_weights`, never in `__init__`. If quantization doesn't run, weight stays `None` (registered parameter) and is handled normally by PyTorch. If quantization does run, the module is frozen at that point. No `.to()` or `_apply` is called after. An override would be dead code.

xfuser/model_executor/quant/aiter_fp8_quantizer.py
```diff
@@ -0,0 +1,342 @@
...
+        # bf16 sentinel (plain attr); forward uses weight_fp8. Do NOT register_parameter("weight",
+        # None) here — the sentinel already occupies `weight` and register_parameter would collide.
+        # Offload: move only the produced fp8 tensors to host. The meta `bias` is streamed
+        # to target_device separately by the normal loader (not a quantized param).
+        if target_device is not None and torch.device(target_device).type != "cuda":
+            module.weight_fp8 = nn.Parameter(
+                module.weight_fp8.data.to(target_device), requires_grad=False
+            )
+            module.weight_scale = module.weight_scale.to(target_device)
+
```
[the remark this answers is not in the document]
pds-amd: The sentinel is 0-element and only exists when `weight_fp8` was also set by `load_and_quantize_weights`. `forward()` always resolves to `weight_fp8`. The `self.weight` fallback only fires when `weight_fp8` is absent, which means the sentinel was never installed either. The sentinel's device is unreachable in any real code path. Confirmed by a successful benchmark sweep at this commit.

xfuser/core/utils/checkpoint_io.py
```diff
@@ -0,0 +1,121 @@
...
+    if idx_path is not None:
+        with open(idx_path) as f:
+            key_to_file = json.load(f)["weight_map"]
+        file_local: dict[str, str] = {}
+        for key, fname in key_to_file.items():
+            if fname not in file_local:
+                file_local[fname] = hf_hub_download(model_name, f"{subfolder}/{fname}")
+            weight_map[key] = file_local[fname]
+    else:
+        single = hf_hub_download(
```
jcaraban: 🤖 This call isn't guarded, so a local model directory raises before the safetensors read. The index download above is wrapped, so `idx_path` becomes `None` and control always reaches: `HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/tmp/tmp75zot94b' `

Local checkpoint dirs are a normal deployment shape, and this kills the whole `--memory_efficient_sharding` transformer self-fill for them. Suggest resolving through `os.path.isdir(model_name)` first (glob the subfolder for `*.safetensors` / `*.index.json`) and only falling through to the Hub otherwise. Same treatment needed in `component_shard_paths` — it swallows the error there, so page-cache dropping just silently no-ops on local paths.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -33,6 +34,7 @@
     rgetattr,
 )
 
+from xfuser.model_executor.quant import AiterFp8BlockScaleConfig  # noqa: F401  (import registers the quantizer)
```
jcaraban: 🤖 This import chain reaches `aiter_fp8_quantizer.py:218`, `from transformers.core_model_loading import ConversionOps`, at module scope. That module doesn't exist in the `transformers>=4.39.1` floor declared in `setup.py`, and `base_model` is imported by every runner module — so an older transformers is a hard ImportError for the whole runner package, not a degraded feature. The `runner_models/__init__.py` guard added in #737 only re-raises non-`diffusers` ImportErrors, so it won't surface as a friendly message either.

Two separable concerns:
- The transformers-side quantizer should be imported lazily (inside `_te_pipeline_quant_config`, which is already gated on `_aiter_fp8_active()`), or the import wrapped so only the TE half is unavailable.
- `register_aiter_fp8_quantizers()` runs at `aiter_fp8_quantizer` import time (line 342), mutating diffusers' and transformers' global `AUTO_QUANTIZER_MAPPING` as an import side effect on every platform. Registering from the same gated call site would keep that scoped to runs that actually use it.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -275,6 +296,151 @@ def initialize(self, input_args: dict) -> None:
...
+                component: AiterFp8BlockScaleTEConfig(target_modules=targets)
+                for component, targets in component_targets.items()
+            },
+        )
+
+    def _memory_efficient_fsdp_load(self) -> bool:
+        """True when the memory-efficient sharded (meta-init + rank0-broadcast) load path is on."""
+        return bool(self.config.memory_efficient_sharding and self.config.fully_shard_degree > 1)
+
+    def _replicated_broadcast_load(self) -> bool:
```
jcaraban, line 462: 🤖 This turns on automatically for every RDNA4 multi-GPU run that isn't using FSDP/pipefusion/TP — there's no way for a user to opt out if it misbehaves. Given the failure mode of a collective bug on this path is a hang rather than an exception, could we gate it behind a CLI flag (or at minimum an `XFUSER_*` env kill-switch) until it's had some mileage? The docstring already explains the win clearly; it just shouldn't be the only option.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -742,27 +949,53 @@ def _shard_model_with_fsdp(self) -> None:
...
                 wrap_attrs = strategy.get("wrap_attrs", [])
                 dtype = strategy.get("dtype", None)
                 offload_policy = strategy.get("offload_policy", None)
-                quantize_fn = self._build_fsdp_quantize_fn(component_name, wrap_attrs, fs_local_rank)
-                reshard_after_forward = self.config.reshard_after_forward
+                # A meta component was built on-config to avoid a full bf16 copy per rank. Two meta
+                # paths: the transformer self-fills each block from disk on every rank (never full
+                # anywhere, quantized per block), while text encoders are filled by a rank0
+                # broadcast (no per-block quantize; the filled TE stays bf16/streamed-fp8 on rank0).
+                is_meta = any(p.is_meta for p in component.parameters())
```
jcaraban: 🤖 This is a rank-local predicate selecting a collective branch. Combined with `meta_load.py:110`'s `except Exception: return None` fallback in `build_meta_component`, a single rank hitting a transient HF/filesystem error ends up with real params here while its peers are meta — so it takes `meta_init=False, load_block_fn=None` and skips `broadcast_load`, while everyone else blocks in a broadcast. That's a hang, not a crash.

`meta_te_kwargs_replicated` deliberately raises instead of falling back for this exact reason, so the asymmetry looks unintentional. Either all-reduce the decision the way `_all_ranks_loaded_real` already does, or let `build_meta_component` propagate and fail every rank.

xfuser/model_executor/models/runner_models/meta_load.py
```diff
@@ -0,0 +1,596 @@
...
+                component = cls._from_config(config)
+            # _from_config defaults to fp32; align meta params to bf16 (dtype-only .to is legal on
+            # meta) so their DTensor dtype matches the broadcast source.
+            component = component.to(torch.bfloat16)
+            if fp8 and self.model._component_wants_fp8(component_name):
+                self._swap_meta_te_to_fp8(
+                    component, self.model._fp8_targets_for_component(component_name)
+                )
+            return component
+        except Exception as e:
```
jcaraban: 🤖 Two issues with the breadth here: it masks genuine bugs in `build_meta_component` (a typo'd config key looks identical to a network blip), and the fallback isn't collective — see the note on `base_model.py:962`. Narrowing to the expected failure set (`OSError`, `EnvironmentError`, `KeyError` on `model_index.json`) would at least make real bugs visible.

xfuser/model_executor/models/runner_models/z_image.py
```diff
@@ -69,13 +69,13 @@ class xFuserZImageModel(xFuserModel):
                 "wrap_attrs": ["noise_refiner", "context_refiner", "layers"],
             },
             "text_encoder": {
-                "wrap_attrs": ["model.language_model.layers"],
+                "wrap_attrs": ["layers"],
             },
         },
-        fp8_gemm_module_list=["transformer.layers", "transformer.noise_refiner", "transformer.context_refiner"],
+        fp8_gemm_module_list=["transformer.layers", "transformer.noise_refiner", "transformer.context_refiner", "text_encoder.layers"],
```
jcaraban: 🤖 This quantizes the Z-Image text encoder to FP8 by default whenever `--use_fp8_gemms` is set — an output-quality change bundled into a memory PR. The `wrap_attrs` correction on line 72 is clearly a bug fix and I'd merge that on its own, but the `fp8_gemm_module_list` addition deserves quality numbers (or its own flag) rather than riding along. Same for line 152.

jcaraban:
Good work! 🤖 Four things to note before merge, and smaller items inline.

**1. The self-fill path breaks on local model directories.** `resolve_checkpoint_weight_map` calls `hf_hub_download(model_name, ...)`. The index lookup is wrapped in `try/except`, but the single-file fallback is not. Reproduced:

```
HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/tmp/tmp75zot94b'
```

Any `--memory_efficient_sharding` run against a local checkpoint directory dies here. Needs an `os.path.isdir(model_name)` branch (or diffusers' `_get_model_file`, which already handles both).

**2. `base_model.py` now hard-imports a transformers-5 module at module scope.** The `# noqa: F401 (import registers the quantizer)` line transitively executes `from transformers.core_model_loading import ConversionOps`. `setup.py` declares `transformers>=4.39.1`, where that module does not exist. Every runner module imports `base_model`, so an older transformers takes down the entire runner package — and #737's `runner_models/__init__.py` guard only re-raises non-`diffusers` ImportErrors, so it won't be masked into a friendly message either. It also means non-RDNA4 users pay the import cost and get their global `AUTO_QUANTIZER_MAPPING` mutated for a code path they can never take.

**3. `_replicated_broadcast_load()` is auto-on with no escape hatch.** It returns True for *any* RDNA4 multi-GPU run without FSDP/pipefusion/TP, silently routing every such run through a brand-new rank0-load + GPU→GPU broadcast path. Given the blast radius of a collective bug here (hang, not crash), I'd want this behind a flag or at least an env opt-out for a release or two.

**4. Rank-asymmetric fallback can deadlock.** `build_meta_component` swallows `Exception` and returns `None` → caller loads real weights. Downstream, `_shard_model_with_fsdp` picks its collective branch with a *local* predicate (`any(p.is_meta ...)`). If one rank hits a transient HF error and falls back, ranks disagree on `meta_init` / `load_block_fn` / `broadcast_load`, and you get mismatched NCCL calls — a silent hang rather than a traceback. `meta_te_kwargs_replicated` already raises instead of falling back for exactly this reason; the two should agree.

**On the Gemini review:** its first finding (`_ckpt_key`'s `getattr` walk raising `AttributeError` on `nn.ModuleList` / `nn.Sequential` index segments) is **wrong** — `nn.Module.__getattr__` resolves from `_modules`, which is keyed by the string index. Verified:

```
getattr(ModuleList, "0") -> Linear
getattr(Sequential, "0") -> Linear
walk 'blocks.0.weight' -> (4, 4)
```

Please don't "fix" that one. Its `.contiguous()` finding is valid but narrow — see inline.

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -7,12 +9,50 @@
...
+        gemm_a8w8_blockscale_preshuffle,
+    )
+    _HAS_PRESHUFFLE = True
+except ImportError:
+    _HAS_PRESHUFFLE = False  # older AITER without preshuffle blockscale GEMM
+
 _FP8_BLOCK = 128
+_PRESHUFFLE_LAYOUT = (16, 16)
+
+# Preshuffle blockscale GEMM measured ~25% SLOWER than plain on gfx1201/RDNA4 (klein 2048^2:
```
avjves: nit: this comments seems like it probably shouldn't be part of the codebase?

xfuser/model_executor/layers/fp8_linear.py
```diff
@@ -47,23 +116,63 @@ def _(
     return torch.empty(M, N, dtype=torch.bfloat16, device=x.device)
 
 
+@torch.library.custom_op("mylib::fp8_blockscale_gemm_preshuffle", mutates_args=())
```
avjves: nit: mylib -> xfuser?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -249,6 +251,25 @@ def _customize_settings(self, config: xFuserArgs) -> None:
     def _update_model_settings(self, config: xFuserArgs) -> None:
         if config.use_fp4_gemms:
             self._apply_fp8_override_cli_from_config(config)
+        self._gate_te_fp8_to_rdna4()
+
+    def _gate_te_fp8_to_rdna4(self) -> None:
```
avjves: I think this is dangerous - the text encoders are explictily set to be part of the FP8 GEMM list and no CLI argument or equivalent user specified action acts as the decider whether to have them there or not. Having it just be for the RDNA4 would be a big side effect. I believe this would/could fool both humans and AI reading the codebase. If we don't want to run these for non-RDNA4, we should gate them explicitly. For example, make the gemm list be similar to FSDP and "enable" it there via an option.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -249,6 +251,25 @@ def _customize_settings(self, config: xFuserArgs) -> None:
...
+        after _customize_settings (which may rebuild the list, e.g. Wan2.2 dual), so the
+        gate covers runtime-assigned lists too.
+        """
+        if _is_hip() and PACKAGES_CHECKER._on_rdna4():
+            return
+        lst = self.settings.fp8_gemm_module_list
+        if not lst:
+            return
+        self.settings.fp8_gemm_module_list = [
+            m for m in lst if m.split(".", 1)[0].startswith("transformer")
```
avjves: This check would work for now (and for the foreseeable future), but could be we _would_ want to add something that doesn't start with `transformer` there. This would be a head-scratcher if it gates them to RDNA only.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -275,6 +296,151 @@ def initialize(self, input_args: dict) -> None:
                 compile_input_args["prompt"] = compile_input_args["prompt"][: self.config.batch_size]
             self._compile_model(compile_input_args)
 
+    def _aiter_fp8_active(self) -> bool:
```
avjves: nit: this is a very clear utility function that should probably be in the utils file? It's a bit of a hazy line what goes there and what doesn't, but AITER specific utility doesn't seem like main runner function :)

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -275,6 +296,151 @@ def initialize(self, input_args: dict) -> None:
...
+        """True when RDNA4 AITER FP8 quantization applies (fp8 gemms requested + supported)."""
+        return bool(self.config.use_fp8_gemms and _use_aiter_fp8_rdna4())
+
+    def _fp8_stream_quant_config(
```
avjves: Could this also be hoisted to utils? We could remove the `AiterFp8BlockScaleConfig` import as well from this file as well. I'm just trying to keep this generic when possible :)

avjves: There seems to be plently of FP8 Streaming related functionality here - could they all be handled elsewhere?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -275,6 +296,151 @@ def initialize(self, input_args: dict) -> None:
...
+    def _memory_efficient_fsdp_load(self) -> bool:
+        """True when the memory-efficient sharded (meta-init + rank0-broadcast) load path is on."""
+        return bool(self.config.memory_efficient_sharding and self.config.fully_shard_degree > 1)
+
+    def _replicated_broadcast_load(self) -> bool:
+        """Auto-on replicated multi-GPU load: the model fits one GPU and is replicated across ranks
+        (pure sequence/CFG/data parallel), so every rank would otherwise from_pretrained a full CPU
+        copy -> host RAM = N x model -> cgroup OOM. Instead rank0 loads real weights to GPU, peers
+        build on meta and receive every param/buffer via a GPU->GPU broadcast (host peak = 1x).
+        RDNA4-only. Excludes weight-splitting parallelism (FSDP/pipefusion/TP), where per-rank
```
avjves: Do these need to be RDNA4 only? I get that that's where the usage is currently, but could we instead turn it on via cli argument? And if we want it on by default then maybe turn on that CLI argument by default (with logging!) on RDNA4. This would keep the logic tied to the behavior wanted, not the hardware.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -275,6 +296,151 @@ def initialize(self, input_args: dict) -> None:
...
+        quantizes per block from fp8_gemm_module_list, so stream_quant only gates the non-meta config.
+        """
+        if self._memory_efficient_fsdp_load():
+            return self._sharder.build_meta_transformer(wrapper_cls, subfolder, init_kwargs)
+        # Replicated broadcast load: build on meta on ALL ranks. Weights are streamed per block from
+        # disk on rank0 and broadcast GPU->GPU, then fp8-quantized per block (broadcast_fill_replicated),
+        # so the full bf16 transformer never materializes on host or any single GPU.
+        if self._replicated_broadcast_load():
+            return self._sharder.build_meta_transformer(wrapper_cls, subfolder, init_kwargs)
+        return wrapper_cls.from_pretrained(
```
avjves: Does this work with the custom transformer stuff we have for the sparse Hunyuan models?

avjves (changes requested):
Thanks! This helps a lot on low memory systems :)
In general I think having these custom build methods for transformers, text encoders, generators etc is a good change - we can now add customization to how it's done. I left some comments while reading, but I think the main things that worry me / I think could be improved are these:
- This is a sizeable change that touches almost the whole codebase. Did you run the standard suite for any regressions (bf16, not RDNA)?
- The FP8 loading is ballooning the base model a lot. Could these be hoisted to their own files?
- A lot of these are now gated to only work on RDNA4. I get that it's the current main use case, but I find it odd to link certain behaviors directly to HW. I think we should only gate functionality to RDNA4 if it only works on RDNA4. If we want the default behavior to be certain on RDNA4, we could enable some other flags / cli arguments by default on that HW. This would allow us to test this on other machines as well.
- There's now a alot AITER FP8 comments infesting the codebase in very generic places - could these be removed? 🫣 
- The FP8 TE to only work on RDNA seems a bit hacky. If we list the text encoder in the list of elements to quantize, we should do it on all systems, unless somehow explicitly disabled.
- There's now a lot of AITER specific functionality in the base model. Any chance this could be reduced?
- Not all models call the custom build methods - why?
- +1 AI comments above

pds-amd:
Moving to draft while I refactor some of this to make it more general

pds-amd:
Superceded by https://github.com/xdit-project/xDiT/pull/751

PR #740, opened 2026-07-16
[VAE] adds --vae_tile_size, fixes config validation and expands DistVAE coverage

pds-amd:
## What

Adds `--vae_tile_size`, one knob that narrows the VAE's tile window on every runner model, plus a guard that tells a failed VAE decode which flag would fix it.

## Why

A VAE's tile window defaults to something near its training resolution and never shrinks. Decoding above that resolution can therefore run out of memory with `--enable_tiling` already on, because it is the tile itself that no longer fits. The only way to narrow the window was to set diffusers attributes on the VAE by hand, and which attributes exist depends on the VAE class.

Chasing that turned up flags already failing quietly. `--enable_tiling` and `--enable_slicing` were doing nothing on models whose capability tables had lost the flags, and on VAE classes the installed diffusers cannot tile at all, because diffusers hands every autoencoder those methods through a mixin whether or not the class implements them.

## Coverage

`--vae_tile_size` takes one edge in output pixels and rescales whichever tiling attributes the VAE class carries, keeping their proportions. Those attributes come in three shapes across the seven VAE classes the runner models load, which is all of them:

| VAE class | attributes rescaled | reached by |
| --- | --- | --- |
| `AutoencoderKL` | `tile_sample_min_size` with `tile_latent_min_size` | SD3, Z-Image, Flux.1 |
| `AutoencoderKLFlux2` | same pair | Flux.2, Ideogram 4 |
| `AutoencoderKLWan` | `tile_sample_min_*` with `tile_sample_stride_*` | Wan 2.1/2.2, CausalWan, Cosmos3, LingBot-Video |
| `AutoencoderKLQwenImage` | same pair | Qwen-Image, Krea 2 |
| `AutoencoderKLHunyuanVideo` | same pair | HunyuanVideo |
| `AutoencoderKLLTX2Video` | same pair | LTX-2, LTX-2.3 |
| `AutoencoderKLHunyuanVideo15` | `tile_sample_min_*` with `tile_latent_min_*` | HunyuanVideo 1.5 |

Overlap factors are read but never rescaled, being fractions of a window that is itself changing. They do constrain which windows are legal: a window whose overlap lands on a fractional latent size is not offered. Frame tiling is left alone throughout, having nothing to do with a spatial tile edge.

A size the attributes cannot express exactly snaps down to the nearest window that works, rather than rounding into a fractional latent size the decoder would fail on. A size above the VAE's own window is declined, since widening it raises peak memory instead of lowering it. Both cases log what they did.

## Behaviour worth knowing

**It turns tiling on by itself.** LTX-2.3 tiles its stage-2 VAE at load, so requiring `--enable_tiling` alongside would mean tiling every stage just to reach the one that ran out of memory.

**Staged models get both VAEs.** A model that decodes in stages loads a second pipeline with its own VAE, and the later stage decodes at full resolution. All VAE options now reach both, which also removes the two `_enable_options` overrides LTX carried for this.

**Parallel VAE reaches five more models, and the rest are listed below.** Making the capability check real turned `--use_parallel_vae` from silently ignored into refused on models that never implemented it. Five of those decode through a VAE DistVAE already has an adapter for, so they are now wired up properly rather than left refusing: Z-Image, Z-Image-Turbo and SD3.5 through `AutoencoderKL`, CausalWan and Wan2.1-VACE through `AutoencoderKLWan`.

Still outstanding, because their decoders need an adapter written in [DistVAE](https://github.com/xdit-project/DistVAE) rather than a declaration here:

| Models | VAE | Blocks with no adapter |
| --- | --- | --- |
| Qwen-Image, Qwen-Image-Edit, Krea-2-Raw, Krea-2-Turbo | `AutoencoderKLQwenImage` | `QwenImageUpBlock`, and no `conv_norm_out` |
| HunyuanVideo | `AutoencoderKLHunyuanVideo` | `HunyuanVideoUpBlock3D` |
| HunyuanVideo 1.5 and its distilled variants | `AutoencoderKLHunyuanVideo15` | `HunyuanVideo15UpBlock3D` |
| LTX-2, LTX-2.3 | `AutoencoderKLLTX2Video` | `LTX2VideoUpBlock3d` |

Those refuse the flag with a message pointing at `--vae_tile_size`, which addresses the same memory pressure without the latency win. Encoder sharding is thinner still: DistVAE ships one encoder adapter, for Wan, so `--use_parallel_vae_encoder` remains limited to Wan I2V and Cosmos 3.

**Wiring parallel VAE is now one call, not a sixth copy.** Five runner files each carried their own module-level `_setup_parallel_vae`, so enabling a sixth model meant copying one of them and knowing which DistVAE adapter its decoder blocks call for by reading someone else's. The VAE answers that itself, from the blocks its decoder is built out of, so `xFuserModel._setup_parallel_vae` now does the whole job and the five copies are gone.

What looked model-specific in them was a function of the VAE's config. Both remaining differences come from `patch_size`: the encoder adapter shards by the encoder's own downsampling, so a VAE that patches has to divide that factor out, and the decoder adapter's patchify assumes no patching because Wan does none. Deriving both reproduces every value the copies passed, Cosmos 3's 16 over 2 included, and does nothing for a VAE that does not patch. Only a single factor counts as patching here, since `patch_size` is not one thing across VAE classes: Flux 2 declares `(2, 2)` for the pixel unshuffle at its boundary, which is not a factor either adapter takes.

That means a missing DistVAE adapter now raises where four of the five copies logged and carried on sharding nothing, or sharding only the decoder when the encoder was wanted too. Refusing up front also replaces an assertion thrown from inside a half-built replacement decoder with an error that names the flag.

**It composes with `--use_parallel_vae`, and is checked up front.** The two divide the same axis: diffusers hands the decoder one tile, then DistVAE splits that tile's latent rows across the VAE group. Below one row per rank, the split yields fewer patches than there are ranks and the surplus ranks index off the end without reporting anything. A window that thin is now refused, naming the smallest window that gives every rank a row.

**LTX-2 wants a wide window.** It compresses space hard enough that a narrow window leaves very few latent rows per tile, and a tile can arrive thinner than a convolution's own padding, which torch reports from deep inside the decoder without mentioning tiles. Whether it happens depends on the output size, so it cannot be caught when the window is set. That failure is now prefaced with the window that caused it, and a larger window can fail where a smaller one works, so the message says to try another size rather than only to go smaller.

**Version floors.** Tiling and slicing arrived class by class across diffusers releases, Wan's tiling in 0.34, one past the `diffusers>=0.33.0` floor in setup.py. An install that satisfies the floor can still be short of the class a model loads, so both flags now raise instead of silently doing nothing.

## Fixes included

- Wan 2.2 distilled I2V, Qwen-Image and Z-Image-Turbo had lost `enable_tiling` and `enable_slicing` from their capability tables. `ModelCapabilities` is frozen, so re-declaring the block without a flag reads as `False`, and `--enable_tiling` was refused on three models that support it.
- The capability check tested every int-valued config with `value > 1`, written for degrees like `ulysses_degree`. `bool` subclasses `int`, so boolean flags took that branch, `True > 1` is `False`, and no boolean capability was ever refused: `--use_parallel_vae` on a model that cannot do it ran as a silent no-op.
- Making that check real exposed CausalWan declaring `use_fp8_gemms=False` in the same class body that lists `fp8_gemm_module_list=["transformer.blocks"]`. The declaration was the wrong half. The declarations this change makes load-bearing were audited against the code reading them, including both hybrid schedule flags, which are declared exactly where `_calculate_hybrid_attention_step_multiplier` accounts for the model's steps.
- One of the two `--enable_slicing` definitions described tiling.

## Structure

The diffusers VAE knowledge, which attributes each class carries and how they scale, lives in `xfuser/core/utils/vae_tiling.py` and reads no xDiT config. Its counterpart `vae_parallel.py` holds the DistVAE half: which adapter a decoder's blocks call for, and the numbers an adapter has to be told about the VAE. Neither reads xDiT config. The runner keeps the policy: what to do with a size that does not fit, what to say when a decode fails.

## Test plan

`tests/core/test_vae_tiling.py` and `tests/core/test_vae_parallel.py`, 37 tests, no downloads. Each of the seven VAE classes is built tiny, resized through the same planning the flag uses, and decoded on CPU, asserting the output matches an untiled decode's size, so a window that only looks valid still has to survive a real decode. The rest covers the three attribute shapes, snapping, the fractional-latent rejection, latent-row counting for the parallel-VAE check, and the padding-error matcher against verbatim text from `AutoencoderKLLTX2Video`.

The parallel-VAE tests pin which adapter each VAE class needs, so a decoder reworked upstream breaks the mapping loudly instead of at someone's first multi-GPU decode, and check that an unshardable VAE is refused before its decode is touched. They also pin the encoder scale factor the five deleted copies used to derive by hand, patching divided out, since that number is now shared. Because the real adapters need a process group, a stand-in stands in for them, which puts every VAE class through the config reads that wrapping does rather than only the ones picking an adapter does.

Not covered by these tests: multi-GPU behaviour, which was reasoned about rather than run. The five newly enabled models each want a two-rank decode checked against a single-rank one before this merges. So does one model per adapter among the eight already using parallel VAE, since their wiring moved to the shared method: the decoder is wrapped by the same adapter with the same arguments, but only a real two-rank decode shows it.

### DIFF
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -346,7 +346,11 @@ class xFuserModel(abc.ABC):
                 vae_tiling.require_vae_support(vae, "tiling", tiling_flag)
                 log(f"Enabling VAE tiling on {type(vae).__name__}...")
                 vae.enable_tiling()
+                # Read the VAE's own window before the next line can narrow it: that window is
+                # the batch budget, so tiling smaller costs the same per call as not tiling.
+                budget_elems = vae_tiling.default_tile_area(vae)
                 tile_window = self._apply_vae_tile_size(vae)
+                self._install_vae_tile_batching(vae, budget_elems)
             # Installed either way, so a decode that OOMs with tiling off still says so.
             self._install_vae_decode_guard(vae, tile_window)
 
@@ -1145,6 +1149,22 @@ class xFuserModel(abc.ABC):
              if smallest else
              f", and no size up to this VAE's own {window}px window gives them one each.")
         )
+
+    def _install_vae_tile_batching(self, vae, budget_elems: Optional[int]) -> None:
+        """Decode a tiled VAE's same-shaped tiles together, within one tile's worth of area"""
+        # Every decoder call costs the same whatever the tile, and under --use_parallel_vae that
+        # cost is a round of collectives rather than arithmetic, so a narrow window pays it over
+        # and over. Batching spends the VAE's own tile area per call instead of one tile, which
+        # leaves the default window decoding exactly as it did and stops the smaller windows
+        # paying for their size twice.
+        if budget_elems is None:
+            return
+        batched = vae_tiling.batched_tiled_decode(vae, budget_elems)
+        if batched is None:
+            return
+        vae.tiled_decode = batched
+        log(f"VAE tiled decode will batch tiles up to {budget_elems} latent elements per call "
+            f"on {type(vae).__name__}.")
 
     def _install_vae_decode_guard(self, vae, tile_window: Optional[int] = None) -> None:
         # Point a failed VAE decode at the knob that fixes it. Success path untouched.
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -344,7 +344,9 @@ class xFuserModel(abc.ABC):
         """ Validate if the model supports requested config """
         for key in ModelCapabilities.__annotations__.keys():
             config_value = getattr(config, key, None)  # Some config options might not be set in the CLI, such as support for specific attention backends.
-            if isinstance(config_value, int):
+            # bool subclasses int, so a boolean flag reaching the degree branch below is
+            # tested with True > 1 and never refused.
+            if isinstance(config_value, int) and not isinstance(config_value, bool):
                 if not getattr(self.capabilities, key) and config_value > 1:
                     raise ValueError(f"Model {self.settings.model_name} does not support {key}.")
             else:
--- a/xfuser/model_executor/models/runner_models/base_model.py
+++ b/xfuser/model_executor/models/runner_models/base_model.py
@@ -345,11 +345,13 @@ class xFuserModel(abc.ABC):
                 vae_tiling.require_vae_support(vae, "tiling", tiling_flag)
                 log(f"Enabling VAE tiling on {type(vae).__name__}...")
                 vae.enable_tiling()
-                # Read the VAE's own window before the next line can narrow it: that window is
-                # the batch budget, so tiling smaller costs the same per call as not tiling.
-                budget_elems = vae_tiling.default_tile_area(vae)
+                # Read the VAE's own window before the next line can narrow it: the budget is
+                # derived from both areas, so both have to be read either side of the narrowing.
+                default_area = vae_tiling.tile_latent_area(vae)
                 tile_window = self._apply_vae_tile_size(vae)
-                self._install_vae_tile_batching(vae, budget_elems)
+                self._install_vae_tile_batching(
+                    vae, default_area, vae_tiling.tile_latent_area(vae)
+                )
             # Installed either way, so a decode that OOMs with tiling off still says so.
             self._install_vae_decode_guard(vae, tile_window)
 
@@ -1149,13 +1151,15 @@ class xFuserModel(abc.ABC):
              f", and no size up to this VAE's own {window}px window gives them one each.")
         )
 
-    def _install_vae_tile_batching(self, vae, budget_elems: Optional[int]) -> None:
-        """Decode a tiled VAE's same-shaped tiles together, within one tile's worth of area"""
+    def _install_vae_tile_batching(
+        self, vae, default_area: Optional[int], tile_area: Optional[int]
+    ) -> None:
+        """Decode a tiled VAE's same-shaped tiles together, within a budget the window sets"""
         # Every decoder call costs the same whatever the tile, and under --use_parallel_vae that
         # cost is a round of collectives rather than arithmetic, so a narrow window pays it over
-        # and over. Batching spends the VAE's own tile area per call instead of one tile, which
-        # leaves the default window decoding exactly as it did and stops the smaller windows
-        # paying for their size twice.
+        # and over. Batching spreads one round over many tiles without taking back all of the
+        # memory the narrow window was asked for; tile_batch_budget is where that balance is set.
+        budget_elems = vae_tiling.tile_batch_budget(default_area, tile_area)
         if budget_elems is None:
             return
         batched = vae_tiling.batched_tiled_decode(vae, budget_elems)
@@ -1163,7 +1167,7 @@ class xFuserModel(abc.ABC):
             return
         vae.tiled_decode = batched
         log(f"VAE tiled decode will batch tiles up to {budget_elems} latent elements per call "
-            f"on {type(vae).__name__}.")
+            f"on {type(vae).__name__} (its own window carries {default_area}).")
 
     def _install_vae_decode_guard(self, vae, tile_window: Optional[int] = None) -> None:
         # Point a failed VAE decode at the knob that fixes it. Success path untouched.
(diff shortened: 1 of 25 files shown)


jcaraban:
🤖  One thing to fix before merge.

**The OOM guard recommends flags that the override path then refuses.** `_apply_vae_tile_size_overrides` correctly bails on VAEs without `tile_latent_min_size`, but the guard installs for every VAE and its fallback branch hits exactly those:

```python
latent = getattr(vae, "tile_latent_min_size", None)   # None on video VAEs
...
else:
    rec = "--vae_tile_latent_size 64 --vae_tile_sample_size 512"
```

So a Wan/HunyuanVideo decode OOM tells the user to pass two flags that print `override ignored` on the next run. Please make the fallback either recommend the video-VAE attrs or say the override isn't supported for this VAE.

Coverage against diffusers 0.38.0.dev0 — only 5 of 24 autoencoders expose the attribute this drives:

- **Supported:** `AutoencoderKL`, `AutoencoderKLFlux2`, `AutoencoderKLHunyuanImage`, `AutoencoderKLKVAE`, `AutoencoderTiny`
- **Not supported** (use `tile_sample_min_height/_width/_num_frames`): `AutoencoderKLWan`, `AutoencoderKLQwenImage`, `AutoencoderKLHunyuanVideo`, `AutoencoderKLLTXVideo`, `AutoencoderKLCogVideoX`, `AutoencoderKLCosmos`, `AutoencoderKLMochi`, …

That's a fine scoping call, but it misses Qwen-Image and every video model — the workloads where decode OOM hurts most. Worth stating in the `--help` text, and supporting the video attrs would be a small follow-up.

Non-blocking suggestions:

1. **Install the guard unconditionally**, not just under `--enable_tiling`. The most instructive OOM is the one *without* tiling, where the answer is "pass `--enable_tiling`" — and nothing currently says so.
2. **Derive the downscale from config rather than current state.** `ratio = max(1, round(sample / latent))` reads back values the user may have just set inconsistently, perpetuating a bad ratio. diffusers derives it structurally: `2 ** (len(vae.config.block_out_channels) - 1)`.
3. **Partial overrides warn, then apply anyway.** Since the partner value is derivable from the same expression, `--vae_tile_latent_size 32` alone could set `tile_sample_min_size = 32 * downscale` and log it — better than warning and then producing seams.
4. `raise torch.cuda.OutOfMemoryError(f"{e}\n\n...") from e` embeds the original message *and* chains it, so it prints twice. `from e` is enough.

Verified along the way: `torch.cuda.OutOfMemoryError` is `torch.OutOfMemoryError` (a `RuntimeError` subclass) and constructs fine from a string, and this wrapper composes correctly with the existing `_convert_vae_to_channels_last` decode patch in either order.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -992,6 +998,61 @@ def decode_wrapper(*args, **kwargs):
...
+        # Default tile size tracks the VAE's training sample_size and never shrinks for
+        # above-training-res decodes, so one tile can exceed free VRAM. CLI shrinks it.
+        # Image VAEs (AutoencoderKL) expose tile_latent_min_size/tile_sample_min_size.
+        # Video VAEs key tiling off tile_sample_min_height/_width/_num_frames instead, so
+        # setattr here would take on a dead attribute and silently do nothing.
+        vae = self.pipe.vae
+        want_latent = self.config.vae_tile_latent_size is not None
+        want_sample = self.config.vae_tile_sample_size is not None
+        if (want_latent or want_sample) and not hasattr(vae, "tile_latent_min_size"):
+            log(f"WARNING: --vae_tile_latent_size / --vae_tile_sample_size unsupported for "
```
avjves: Would it not be better to raise an error? In what cases would one want to keep going if the params they used were ignored? Works similarly with all the others via ModelCapabilities

pds-amd: Agreed, changed to a `ValueError` naming the model and VAE class. I also moved the "requires --enable_tiling" case into `_validate_config` so it fails before the model loads, and added a positive-value check.

pds-amd:
Changes have been addressed.

I also noticed three classes had silently lost `enable_tiling`/`enable_slicing` at some point which I have added back in.

pds-amd:
Switching to draft... want to investigate if there's actually a more general way to implement this that will also work for some of the models that use different VAEs (e.g. video) as well

pds-amd:
Superceded by https://github.com/xdit-project/xDiT/pull/750

PR #741, opened 2026-07-16
Unified step caching entrypoint and dbcache presets

pds-amd:
Step caching skips the full transformer forward on timesteps where the output would be nearly identical to the previous step which can result in significant latency improvements especially on consumer hardware.

`main` already has teacache (Flux1) and fbcache (Flux.2) behind a class-registry dispatcher. This PR replaces that with a string-method dispatcher (`--cache_method teacache|fbcache|dbcache`), adds DBCache (from [cache-dit](https://github.com/vipshop/cache-dit)) with some balanced model tuned configs overridable with `--cache_config`.

[DBCache ](https://cache-dit.readthedocs.io/en/latest/user_guide/DBCACHE_DESIGN/) is a hybrid acceleration strategy that balanced the speed/image quality trade off very well and is a nice convenience wrapper for adding performant cache acceleration to xDiT.


jcaraban:
🤖 Two blocking items.

**1. The legacy pipeline path drops the xFuser transformer wrapper when caching is on.** In `_convert_transformer_backbone` the local is rebound to the wrapper:

```python
transformer = wrapper(transformer)
```

but the cache block round-trips through the pipe:

```python
apply_cache(..., pipe=pipe, ...)   # reads getattr(pipe, "transformer")
transformer = pipe.transformer
```

At that point `pipe.transformer` is still the **unwrapped** module — `__init__` only assigns `pipeline.transformer` from this function's return value. So the cache is applied to the unwrapped transformer and the return value replaces the wrapper.

Reachable: wrapping fires when any of SP/PP/CFG/TP > 1 or fast-attn is on, while the in-tree guard only requires `PP == CFG == TP == 1`. So **SP > 1 + `--use_teacache` hits it** — exactly the USP case `adapters/flux.py` advertises as "USP-safe via all_reduce". Passing the local `transformer` into `apply_cache` and returning it, instead of going via `pipe`, fixes this.

**2. `_install_sp_can_cache_sync` fails silently.**

```python
except ImportError:
    return
```

This patches a deep third-party internal (`cache_dit.caching.cache_contexts.cache_manager`) and setup.py pins only `cache-dit>=1.5.0` with no upper bound. If that path moves, the patch silently doesn't install and the rank-divergence NCCL hang returns with no indication — the worst failure mode for a guard that exists to prevent a hang. Please raise (or log an error) and add an upper bound to the extra.

Medium:

3. **An unrelated diffusers-API change is buried in `utils.py`** — the Flux single-block convention switches from concat/split to `encoder, hidden = block(hidden, encoder, ...)`. Correct for current diffusers (`FluxSingleTransformerBlock.forward` returns `(encoder_hidden_states, hidden_states)`), but uncommented, unmentioned in the description, and it silently breaks TeaCache on older diffusers still inside the declared `>=0.33.0` floor. Deserves its own commit plus a note.

4. **`--cache_threshold` is introduced already deprecated.** It doesn't exist on `main`; this PR adds and deprecates it in one change, and registers it only in the second parser block, not `runtime_group` — so the `__post_init__` migration is dead on the main CLI path. Either add it properly or drop the flag and keep only the `__post_init__` handling.

5. **`DeprecationWarning` is filtered by default** outside `__main__`, so `--use_teacache` users will likely see nothing. `UserWarning` or the existing `log()` helper would actually surface.

6. **Unknown `--cache_config` keys warn-and-drop, rank0 only** — a typo silently runs preset defaults. Inconsistent with `_resolve_threshold`, which deliberately fails loudly on bad JSON.

Nits:

- `register_buffer("rel_l1_thresh", ...)` should probably be `persistent=False`; it's a config scalar and now lands in `state_dict()`. (The fix itself is good — the old `.to(get_device(0))` pinned every rank to GPU 0.)
- Two different classes named `DBCacheConfig` (xfuser's bundling dataclass vs cache-dit's real config, passed as `DBCacheConfig: Any`). Renaming the xfuser one would save confusion.
- `_build_adapter` disables `check_forward_pattern` because the cache is applied after `_compile_model`, so blocks are already `OptimizedModule`. Applying the cache before compile would remove the need — any reason for that ordering?

Verified while reviewing: `shard_component` already accepts `forward_prefetch`, so the new `forward_prefetch=not bool(self.config.cache_method)` is safe, and the `diffusers_adapters` → `adapters` rename left no stale imports.

xfuser/config/args.py
```diff
@@ -225,12 +269,33 @@ def add_cli_args(parser: FlexibleArgumentParser):
...
+            help="[Deprecated] Use --cache_method fbcache instead.",
+        )
+        runtime_group.add_argument(
+            "--cache_method",
+            type=str,
+            default=None,
+            choices=["teacache", "fbcache", "dbcache"],
+            help=(
+                "Step caching method. "
+                "teacache: xDiT in-tree (Flux1 only). "
```
avjves: I think we can remove the models from here - we will just forget to update them in the future. As long as the models have proper guards it should be fine.

xfuser/model_executor/cache/adapters/__init__.py
```diff
@@ -0,0 +1,120 @@
+"""
+xDiT step-caching unified entry point.
+
+Implementation files:
+  flux.py:      Flux1 xDiT in-tree TeaCache (USP-native via all_reduce)
```
avjves: These comments can probably be removed as well?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -295,10 +301,36 @@ def _enable_options(self) -> None:
...
+            adapter_config=method_cfg.adapter if method_cfg else None,
+            cache_config=self.config.cache_config,
+            transformer_attr=self.settings.transformer_attr_names[0],
+        )
+        log(f"Step cache applied: method={cache_method}")
 
     def _validate_config(self, config: xFuserArgs) -> None:
         """ Validate if the model supports requested config """
         for key in ModelCapabilities.__annotations__.keys():
+            if key == "supported_cache_methods":
```
avjves: I think the idiomatic way would be to add the "--cache-method" as the ModelCapabilities and the supported cache methods as ModelSettings.

Still need to check below if the cache method is in the supported cache methods, but we can remove this check and capabilities check below.

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -385,6 +428,8 @@ def _load_model(self) -> DiffusionPipeline:
         )
         _load_distilled_weights(transformer,   self.config.distilled_transformer_path)
         _load_distilled_weights(transformer_2, self.config.distilled_transformer_2_path)
+        # Distilled reloads raw bf16 weights via strict load_state_dict, so streaming
```
avjves: Is this a comment from a different PR? 🤔

avjves (commented):
Thanks! Left a couple small comments and ideas :)

avjves (approved):

PR #742, opened 2026-07-17
Add FLUX - AMD Quark quantitation example

amd-xiaoyu12:


pds-amd (commented):
Thanks for this contribution.

I tried running the example on a gfx1201 machine with 32GB VRAM and it OOMs - it looks like there's a hard assumption that the whole model and quantized weights can be resident on the GPU. Sequential CPU offload also failed.

On RDNA4, `--use_fp8_gemms` already dispatches to aiter block-128 (`xFuserFP8BlockScaleLinear`); on MI3xx it uses torchao per-tensor. I'm assuming you're testing this on MI3xx?

Could you please provide some details in your PR description, e.g.

- frame running this against `--use_fp8_gemms`/`--use_fp4_gemms`, and give benchmarks (latency, memory, etc) vs that baseline.
- output images

Please also see the Gemini comments.

avjves:
Thanks!

And to add to @pds-amd's comments, we have generally deprecated the examples in favor of using the model runner. (see [here](https://github.com/xdit-project/xDiT/blob/main/docs/runner/runner.md)). Any chance you could add the support to the model runner instead of a one-off example?

PR #744, opened 2026-07-24
feat: add AITER CK VSA for Wan

LiuYinfeng01:
### Summary

This PR adds a Jenga-compatible block-sparse self-attention path for Wan2.1 using AITER's CK-Tile VSA kernel. It supports 4-way Ulysses sequence parallelism, keeps cross-attention and accuracy-sensitive denoising steps on dense AITER, and exposes mask, scheduling, and TeaCache controls through the xDiT runtime configuration.

### Principle

The Wan VSA path preserves the model's logical token order around distributed attention while presenting spatially local blocks to the sparse kernel:

1. Wan publishes the patch-grid shape `(T, H, W)` and the current denoising-step schedule through `attention_kwargs`.
2. USP performs the normal Ulysses Q/K/V all-to-all. The VSA setup then removes SP padding, de-interleaves the sequence, and applies a sliced-Gilbert permutation to video tokens.
3. Q/K are mean-pooled in 128-token blocks. A Jenga-style probability mask retains the smallest sorted block set whose cumulative probability reaches the configured threshold, subject to the drop-rate-derived `top_k` floor.
4. The dynamic mask is unioned with static Gilbert-neighbor relations and first-frame relations; padded KV blocks are removed.
5. The boolean block mask is converted to AITER's int32 delta-LUT plus valid-block counts and executed by `vsa_sparse_attention`.
6. The output is inverse-permuted, re-interleaved, and returned through USP's output all-to-all.

Step 0 and effective drop rates `<= 0.25` use dense AITER. Cross-attention, calls without Wan grid metadata, and unsupported sparse cases also fall back to dense AITER. This avoids combining xDiT's generic five-step high-precision warmup with Jenga's own dynamic schedule.

### Changes

**AITER VSA backend**
- Added `AttentionBackendType.AITER_VSA` and runtime availability checks for `aiter.ops.jenga_sparse_attention.vsa_sparse_attention`.
- Added Jenga block-mask selection, delta-LUT encoding, CK kernel dispatch, dense fallback, and device/workspace caches.

**Wan and Ulysses integration**
- Added sliced-Gilbert permutation and static-neighbor mask support to the shared Sparge sequence setup/restore path.
- Added per-denoising-step VSA scheduling, Wan patch-grid metadata propagation, inference-state reset, and Ulysses-compatible dispatch.
- Added configurable block size, drop rates, probability threshold, sequence reorder, static-neighbor mask, and first-frame mask.

**Wan TeaCache**
- Added sequence-parallel Wan TeaCache state management, threshold controls, and optional retention-step coefficients.

**Validation tools**
- Added mask/LUT/schedule/Gilbert and CK-vs-masked-dense unit tests.
- Added per-timestep accuracy tracing and paired Dense/VSA video validation benchmarks.

### Tests

```bash
PYTHONPATH=. pytest tests/core/test_vsa_attention.py -v
```

Result: `6 passed`, including the AITER CK VSA kernel comparison against masked dense attention on GPU.

```bash
python3 -m py_compile   xfuser/config/args.py xfuser/config/config.py   xfuser/core/distributed/attention_backend.py   xfuser/core/distributed/runtime_state.py   xfuser/core/sparge_attention/sparge.py   xfuser/core/vsa_attention.py   xfuser/model_executor/models/runner_models/base_model.py   xfuser/model_executor/models/runner_models/wan.py   xfuser/model_executor/models/transformers/transformer_wan.py   benchmarks/wan_vsa_accuracy.py   benchmarks/wan_vsa_video_validation.py   tests/core/test_vsa_attention.py
```

### Benchmark

4× AMD MI308X, Wan2.1-T2V-1.3B, Ulysses=4, 3 prompts × 3 seeds, 50 denoising steps, 480×832, 81 frames, guidance scale 6.0, flow shift 8.0, TeaCache disabled. Dense and VSA each used an independent warmup; the first timing was discarded and two measured iterations were retained per case.

```bash
torchrun --nproc_per_node=4 benchmarks/wan_vsa_video_validation.py   --model-path <WAN_MODEL_PATH>   --output wan-vsa-validation.json   --seeds 0 1 2   --steps 50 --height 480 --width 832 --frames 81   --iterations 3
```

**xDiT end-to-end results**
- Dense: mean `65.367 s`, median `65.368 s`, stddev `0.019 s`
- AITER VSA: mean `63.153 s`, median `63.260 s`, stddev `0.265 s`
- Dense / VSA: mean `1.0351×`, median `1.0331×`, speedup stddev `0.0043`
- Dense-vs-VSA quality: PSNR `30.039`, SSIM `0.9469`, LPIPS-Alex `0.02113`

A native Wan implementation using the same weights, kernels, generation settings, and warmup protocol measured `1.0318×`. The relative speedup difference is `0.318%`, within the 2% engineering-consistency target. The remaining approximately 6.38-second absolute framework gap is nearly identical for Dense and VSA and therefore does not account for the sparse-attention speedup.

### Limitations

- Requires an AITER build containing the CK-Tile `jenga_sparse_attention` operator.
- VSA currently supports non-causal Wan self-attention with block size 128; cross-attention remains dense.
- Ring parallelism greater than one is not supported by this backend.

### DIFF
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -130,6 +130,8 @@ class xFuserArgs:
     coco_path: Optional[str] = None
     use_cache: bool = False
     use_teacache: bool = False
+    wan_teacache_thresh: float = 0.2
+    wan_teacache_use_ret_steps: bool = False
     use_fbcache: bool = False
     # Other arguments
     use_fp8_t5_encoder: bool = False
@@ -176,6 +178,16 @@ class xFuserArgs:
     spargeattn_simthreshold: float = 0.3
     spargeattn_cdfthreshold: float = 0.92
     use_spargeattn_head_balance: bool = False
+    # AITER CK-Tile VSA attention
+    vsa_block_size: int = 128
+    vsa_top_k: int = 1
+    vsa_top_k_ratio: float = 0.0
+    vsa_drop_rates: Optional[List[float]] = None
+    vsa_calls_per_step: int = 2
+    vsa_prob_threshold: float = 0.9
+    vsa_reorder_sequence: bool = True
+    use_vsa_static_block_mask: bool = True
+    use_vsa_first_frame_mask: bool = True
     # Distilled model weight paths
     distilled_transformer_path: Optional[str] = None
     distilled_transformer_2_path: Optional[str] = None
@@ -225,7 +237,18 @@ class xFuserArgs:
         runtime_group.add_argument(
             "--use_teacache",
             action="store_true",
-            help="Enable teacache to accelerate inference in a single card",
+            help="Enable TeaCache. Wan supports sequence parallel TeaCache.",
+        )
+        runtime_group.add_argument(
+            "--wan_teacache_thresh",
+            type=float,
+            default=0.2,
+            help="Wan TeaCache accumulated relative-L1 threshold.",
+        )
+        runtime_group.add_argument(
+            "--wan_teacache_use_ret_steps",
+            action="store_true",
+            help="Use Wan TeaCache retention-step coefficients and warmup.",
         )
         runtime_group.add_argument(
             "--use_fbcache",
@@ -800,6 +823,64 @@ class xFuserArgs:
                  "effect with ulysses_degree>1 and a Sparge attention backend.",
         )
         parser.add_argument(
+            "--vsa_block_size",
+            type=int,
+            default=128,
+            help="Query/KV block size for the AITER CK-Tile VSA backend.",
+        )
+        parser.add_argument(
+            "--vsa_top_k",
+            type=int,
+            default=1,
+            help="Minimum KV blocks selected per query block by AITER VSA.",
+        )
+        parser.add_argument(
+            "--vsa_top_k_ratio",
+            type=float,
+            default=0.0,
+            help="Minimum selected KV-block fraction for AITER VSA. "
+                 "For Jenga drop rate r, set this to 1-r.",
+        )
+        parser.add_argument(
+            "--vsa_drop_rates",
+            type=float,
+            nargs="+",
+            default=None,
+            help="Enable Jenga's per-step sparse schedule. One value applies "
+                 "to all steps; two values switch after the midpoint. "
+                 "Drop rates <=0.25 use dense AITER attention.",
+        )
+        parser.add_argument(
+            "--vsa_calls_per_step",
+            type=int,
+            default=2,
+            help="Legacy override; Wan infers calls per step from CFG mode.",
+        )
+        parser.add_argument(
+            "--vsa_prob_threshold",
+            type=float,
+            default=0.9,
+            help="Jenga cumulative probability threshold for AITER VSA.",
+        )
+        parser.add_argument(
+            "--vsa_reorder_sequence",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help="Reorder Wan video tokens with the sliced Gilbert curve before VSA.",
+        )
+        parser.add_argument(
+            "--use_vsa_static_block_mask",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help="Union Gilbert physical-neighbor blocks into the dynamic VSA mask.",
+        )
+        parser.add_argument(
+            "--use_vsa_first_frame_mask",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help="Preserve first-frame block relations in the VSA mask.",
+        )
+        parser.add_argument(
             "--distilled_transformer_path",
             type=nullable_str,
             default=None,
@@ -893,6 +974,10 @@ class xFuserArgs:
             use_parallel_vae=self.use_parallel_vae,
             use_torch_compile=self.use_torch_compile,
             use_onediff=self.use_onediff,
+            use_teacache=self.use_teacache,
+            wan_teacache_thresh=self.wan_teacache_thresh,
+            wan_teacache_use_ret_steps=self.wan_teacache_use_ret_steps,
+            use_fbcache=self.use_fbcache,
             # use_profiler=self.use_profiler,
             use_fp8_t5_encoder=self.use_fp8_t5_encoder,
             attention_backend=self.attention_backend,
@@ -902,6 +987,15 @@ class xFuserArgs:
             spargeattn_simthreshold=self.spargeattn_simthreshold,
             spargeattn_cdfthreshold=self.spargeattn_cdfthreshold,
             use_spargeattn_head_balance=self.use_spargeattn_head_balance,
+            vsa_block_size=self.vsa_block_size,
+            vsa_top_k=self.vsa_top_k,
+            vsa_top_k_ratio=self.vsa_top_k_ratio,
+            vsa_drop_rates=self.vsa_drop_rates,
+            vsa_calls_per_step=self.vsa_calls_per_step,
+            vsa_prob_threshold=self.vsa_prob_threshold,
+            vsa_reorder_sequence=self.vsa_reorder_sequence,
+            use_vsa_static_block_mask=self.use_vsa_static_block_mask,
+            use_vsa_first_frame_mask=self.use_vsa_first_frame_mask,
         )
 
         parallel_config = ParallelConfig(
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -957,6 +957,7 @@ class xFuserArgs:
         runtime_config = RuntimeConfig(
             warmup_steps=self.warmup_steps,
             # use_cuda_graph=self.use_cuda_graph,
+            use_hybrid_attn_schedule=self.use_hybrid_attn_schedule,
             use_parallel_vae=self.use_parallel_vae,
             use_torch_compile=self.use_torch_compile,
             use_onediff=self.use_onediff,
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -130,8 +130,6 @@ class xFuserArgs:
     coco_path: Optional[str] = None
     use_cache: bool = False
     use_teacache: bool = False
-    wan_teacache_thresh: float = 0.2
-    wan_teacache_use_ret_steps: bool = False
     use_fbcache: bool = False
     # Other arguments
     use_fp8_t5_encoder: bool = False
@@ -183,11 +181,11 @@ class xFuserArgs:
     vsa_top_k: int = 1
     vsa_top_k_ratio: float = 0.0
     vsa_drop_rates: Optional[List[float]] = None
-    vsa_calls_per_step: int = 2
     vsa_prob_threshold: float = 0.9
     vsa_reorder_sequence: bool = True
     use_vsa_static_block_mask: bool = True
     use_vsa_first_frame_mask: bool = True
+    vsa_collect_density: bool = False
     # Distilled model weight paths
     distilled_transformer_path: Optional[str] = None
     distilled_transformer_2_path: Optional[str] = None
@@ -237,18 +235,7 @@ class xFuserArgs:
         runtime_group.add_argument(
             "--use_teacache",
             action="store_true",
-            help="Enable TeaCache. Wan supports sequence parallel TeaCache.",
-        )
-        runtime_group.add_argument(
-            "--wan_teacache_thresh",
-            type=float,
-            default=0.2,
-            help="Wan TeaCache accumulated relative-L1 threshold.",
-        )
-        runtime_group.add_argument(
-            "--wan_teacache_use_ret_steps",
-            action="store_true",
-            help="Use Wan TeaCache retention-step coefficients and warmup.",
+            help="Enable teacache to accelerate inference in a single card",
         )
         runtime_group.add_argument(
             "--use_fbcache",
@@ -851,12 +838,6 @@ class xFuserArgs:
                  "Drop rates <=0.25 use dense AITER attention.",
         )
         parser.add_argument(
-            "--vsa_calls_per_step",
-            type=int,
-            default=2,
-            help="Legacy override; Wan infers calls per step from CFG mode.",
-        )
-        parser.add_argument(
             "--vsa_prob_threshold",
             type=float,
             default=0.9,
@@ -879,6 +860,11 @@ class xFuserArgs:
             action=argparse.BooleanOptionalAction,
             default=True,
             help="Preserve first-frame block relations in the VSA mask.",
+        )
+        parser.add_argument(
+            "--vsa_collect_density",
+            action="store_true",
+            help="Record the selected VSA block density for profiling.",
         )
         parser.add_argument(
             "--distilled_transformer_path",
@@ -974,10 +960,6 @@ class xFuserArgs:
             use_parallel_vae=self.use_parallel_vae,
             use_torch_compile=self.use_torch_compile,
             use_onediff=self.use_onediff,
-            use_teacache=self.use_teacache,
-            wan_teacache_thresh=self.wan_teacache_thresh,
-            wan_teacache_use_ret_steps=self.wan_teacache_use_ret_steps,
-            use_fbcache=self.use_fbcache,
             # use_profiler=self.use_profiler,
             use_fp8_t5_encoder=self.use_fp8_t5_encoder,
             attention_backend=self.attention_backend,
@@ -991,11 +973,11 @@ class xFuserArgs:
             vsa_top_k=self.vsa_top_k,
             vsa_top_k_ratio=self.vsa_top_k_ratio,
             vsa_drop_rates=self.vsa_drop_rates,
-            vsa_calls_per_step=self.vsa_calls_per_step,
             vsa_prob_threshold=self.vsa_prob_threshold,
             vsa_reorder_sequence=self.vsa_reorder_sequence,
             use_vsa_static_block_mask=self.use_vsa_static_block_mask,
             use_vsa_first_frame_mask=self.use_vsa_first_frame_mask,
+            vsa_collect_density=self.vsa_collect_density,
         )
 
         parallel_config = ParallelConfig(
--- a/xfuser/config/config.py
+++ b/xfuser/config/config.py
@@ -55,6 +55,7 @@ class RuntimeConfig:
     warmup_steps: int = 1
     dtype: torch.dtype = torch.float16
     use_cuda_graph: bool = False
+    use_hybrid_attn_schedule: bool = False
     use_parallel_vae: bool = False
     use_profiler: bool = False
     use_torch_compile: bool = False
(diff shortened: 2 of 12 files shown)


LiuYinfeng01:
Hi @jcaraban, could you please review this AITER CK-Tile VSA integration for Wan when you have time? It includes 4-way Ulysses support, Jenga-compatible mask/LUT generation, Dense/VSA accuracy tests, and 4× MI308X benchmark results. Thank you!

xfuser/config/args.py
```diff
@@ -130,6 +130,8 @@ class xFuserArgs:
     coco_path: Optional[str] = None
     use_cache: bool = False
     use_teacache: bool = False
+    wan_teacache_thresh: float = 0.2
+    wan_teacache_use_ret_steps: bool = False
```
pds-amd: Note: There is an open PR to create a unified step caching entry point (https://github.com/xdit-project/xDiT/pull/741). Depending on which PR lands first, some adaptation should be done so it is used consistently across models.

**The examples, tests and PR don't illustrate the impact of this step caching addition at all.**

xfuser/config/args.py
```diff
@@ -176,6 +178,16 @@ class xFuserArgs:
     spargeattn_simthreshold: float = 0.3
     spargeattn_cdfthreshold: float = 0.92
     use_spargeattn_head_balance: bool = False
+    # AITER CK-Tile VSA attention
+    vsa_block_size: int = 128
+    vsa_top_k: int = 1
+    vsa_top_k_ratio: float = 0.0
+    vsa_drop_rates: Optional[List[float]] = None
+    vsa_calls_per_step: int = 2
+    vsa_prob_threshold: float = 0.9
```
pds-amd, line 184: Why is this defaulted to 0.9 but `_aiter_vsa_attn_call` defaults to 0.8?

`build_jenga_block_mask()` also uses `prob_threshold: float = 0.8`.

These default values should agree. Any direct API use or custom pipelines could silently get an unintended value.

xfuser/config/args.py
```diff
@@ -799,6 +822,64 @@ def add_runner_args(parser: FlexibleArgumentParser):
...
+            default=None,
+            help="Enable Jenga's per-step sparse schedule. One value applies "
+                 "to all steps; two values switch after the midpoint. "
+                 "Drop rates <=0.25 use dense AITER attention.",
+        )
+        parser.add_argument(
+            "--vsa_calls_per_step",
+            type=int,
+            default=2,
+            help="Legacy override; Wan infers calls per step from CFG mode.",
```
pds-amd: This is a new feature? Why is this "legacy"? It's also set to 2 here and 1 elsewhere. If I trace this argument through the code it's not even used since `calls_per_step` is used directly.

xfuser/config/config.py
```diff
@@ -61,6 +61,8 @@ class RuntimeConfig:
     use_onediff: bool = False
     use_fp8_t5_encoder: bool = False
     use_teacache: bool = False
+    wan_teacache_thresh: float = 0.2
```
pds-amd: Similar to the step caching note in args.py - this is too model specific for such a general dataclass.

xfuser/core/distributed/attention_backend.py
```diff
@@ -822,6 +823,104 @@ def _aiter_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=N
...
+            return _aiter_attn_call(
+                query,
+                key,
+                value,
+                dropout_p,
+                is_causal,
+                attention_kwargs,
+            )
+
+    density_sink = attention_kwargs.get("vsa_density_sink")
```
pds-amd: `vsa_density_sink` is never set by any code in the PR; always `None`; the density-sink branch is unreachable and not used in any of the code in this PR.

xfuser/core/distributed/attention_backend.py
```diff
@@ -822,6 +823,104 @@ def _aiter_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=N
...
+    collect_density = bool(
+        density_sink is not None
+        or attention_kwargs.get("vsa_collect_density", False)
```
pds-amd: The only way to enable it is direct object mutation like in `wan_vsa_accuracy.py`:

```
  runner.model.pipe.transformer.attention_kwargs["vsa_collect_density"] = True
```

There's no config field, no CLI flag, no docstring. Anyone who wants sparsity profiling in their own script has to rediscover this by reading kernel source. The other VSA params (`vsa_block_size`, `vsa_prob_threshold`, etc.) all go through `_build_attention_kwargs` so `vsa_collect_density` should too.

xfuser/core/sparge_attention/sparge.py
```diff
@@ -229,11 +258,24 @@ def setup_sparge(
         if static_mask is not None:
             n_iq = (image_len + img_pad) // block_m
             n_ik = (image_len + img_pad) // block_n
-            padded_static = torch.zeros(
-                (n_iq, n_ik), dtype=static_mask.dtype, device=static_mask.device
+            padded_key = (
+                static_mask.data_ptr(),
```
pds-amd: `data_ptr()` encodes memory address, not tensor identity. If the underlying tensor is freed and a new allocation lands at the same address, stale cached results are returned. Use the tensor object itself as the key, or ensure `_STATIC_BLOCK_MASK_CACHE` is documented as the anchor that keeps the address stable.

xfuser/core/vsa_attention.py
```diff
@@ -0,0 +1,367 @@
...
+    """Reproduce Jenga's timestep-dependent self-attention drop rate."""
+    if total_steps <= 0:
+        raise ValueError(f"total_steps must be positive, got {total_steps}")
+    if not drop_rates:
+        raise ValueError("drop_rates must contain at least one value")
+    if any(rate < 0.0 or rate > 1.0 for rate in drop_rates):
+        raise ValueError(f"drop rates must be in [0,1], got {drop_rates}")
+
+    step_index = min(max(int(step_index), 0), total_steps - 1)
+    if len(drop_rates) == 1 or step_index <= 25:
```
pds-amd: What is the reason for a hard boundary at step 25? When total steps <26 the second rate is never applied which will silently break short runs. Why is this not relative to the total number of steps? Does this imply the implementation of Jenga you are adding is only useful when doing exactly 50 steps? This looks like a very important implementation detail.

xfuser/core/vsa_attention.py
```diff
@@ -0,0 +1,367 @@
...
+            int(top_k),
+            int((query.shape[2] // block_size) * (1.0 - drop_rate)),
+        )
+
+    first_frame_blocks = 0
+    if use_first_frame_mask:
+        # Match Jenga's decoupled first-frame rule: divide the padded block
+        # count evenly across temporal slices and keep the first slice local.
+        num_blocks = query.shape[2] // block_size
+        first_frame_blocks = num_blocks // int(thw[0])
```
pds-amd: If `num_blocks` is not evenly divisible by `T`, the first-frame block count is silently wrong. Should document or assert the expected relationship between `num_blocks` and `thw`.

xfuser/core/vsa_attention.py
```diff
@@ -0,0 +1,367 @@
...
+
+    block_mask = build_jenga_block_mask(
```
pds-amd, line 337: Is the absence of any caching intentional?

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -47,18 +47,60 @@
...
+        "vsa_reorder_sequence": config.vsa_reorder_sequence,
+        "use_vsa_static_block_mask": config.use_vsa_static_block_mask,
+        "use_vsa_first_frame_mask": config.use_vsa_first_frame_mask,
+        "use_wan_teacache": config.use_teacache,
+        "wan_teacache_thresh": config.wan_teacache_thresh,
+        "wan_teacache_use_ret_steps": config.wan_teacache_use_ret_steps,
     }
 
 
+def _prepare_wan_inference_run(pipe, config, input_args: dict) -> None:
```
pds-amd: This `_prepare_wan_inference_run` pattern is fragile. Consider moving the call into `_run_timed_pipe` in the base class (or a Wan-specific base class that all Wan runners inherit), not scattered across 5 override bodies.

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -438,6 +490,14 @@ def _run_pipe(self, input_args: dict) -> DiffusionOutput:
 @register_model("Wan2.1-T2V")
 class xFuserWan21T2VModel(xFuserModel):
 
+    def preprocess_args(self, input_args: dict) -> dict:
+        args = super().preprocess_args(input_args)
+        if self.config.attention_backend == "AITER_VSA":
+            # Avoid stacking xDiT's five-step warmup on Jenga's dynamic one.
+            args["num_hybrid_attn_high_precision_steps"] = 0
+            self.config.num_hybrid_attn_high_precision_steps = 0
```
pds-amd: `_setup_hybrid_attn_schedule` should gate on `attention_backend`, not be zeroed out via config mutation implicitly through `preprocess_args`. A one-line guard at the top of `_setup_hybrid_attn_schedule` should be sufficient. This applies above as well.

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -320,7 +384,87 @@ def get_rotary_emb_chunk(freqs, pad_amount):
...
+            if getattr(self.config, "in_channels", 16) != 16:
+                raise NotImplementedError(
+                    "xDiT Wan TeaCache currently supports Wan2.1 T2V only"
+                )
+            use_ret_steps = bool(
+                self.attention_kwargs.get(
+                    "wan_teacache_use_ret_steps", False
+                )
+            )
+            is_1_3b = len(self.blocks) <= 30
```
pds-amd: Block count is an unreliable model identifier. If a future Wan variant has ≤30 blocks but different characteristics, wrong coefficients are silently used. The model ID or in_channels would be safer discriminators.

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -320,7 +384,87 @@ def get_rotary_emb_chunk(freqs, pad_amount):
...
+                          5.87365115e1, -3.15583525e-1]
+                )
+            else:
+                coefficients = (
+                    [2.39676752e3, -1.31110545e3, 2.01331979e2,
+                     -8.29855975, 1.37887774e-1]
+                    if is_1_3b
+                    else [-5784.54975374, 5449.50911966, -1811.16591783,
+                          256.27178429, -13.02252404]
+                )
```
pds-amd: Can you please leave a note here about where these coefficients come from. How were they selected / derived?

pds-amd (changes requested):
Thanks @LiuYinfeng01 - I have done an initial review. Please see the comments I have added.

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -320,7 +384,87 @@ def get_rotary_emb_chunk(freqs, pad_amount):
...
+        use_wan_teacache = (
+            self.attention_kwargs is not None
+            and self.attention_kwargs.get("use_wan_teacache", False)
+            and not torch.is_grad_enabled()
+        )
+        if use_wan_teacache:
+            if getattr(self.config, "in_channels", 16) != 16:
+                raise NotImplementedError(
+                    "xDiT Wan TeaCache currently supports Wan2.1 T2V only"
+                )
```
pds-amd: `NotImplementedError` mid-inference instead of at config validation time.

xfuser/config/args.py
LiuYinfeng01: Fixed in 658baab. The VSA probability threshold now defaults to 0.9 consistently in xFuserArgs, RuntimeConfig, _aiter_vsa_attn_call, aiter_vsa_attention, and build_jenga_block_mask. A unit test also verifies the public defaults.

xfuser/core/vsa_attention.py
LiuYinfeng01: Yes, for the dynamic probability mask: it depends on the current Q/K tensors and therefore cannot be reused across denoising calls. The reusable layout permutations and static-neighbor masks remain cached by setup_sparge; 658baab also replaces data_ptr-based identities with stable layout/device cache keys and documents this distinction.

LiuYinfeng01:
@pds-amd Thank you very much for the thorough and thoughtful review. I really appreciate your detailed feedback—it helped improve the consistency and robustness of the implementation. I’ve addressed all the comments and pushed the fixes. Thanks again for taking the time to review this PR!

LiuYinfeng01:
@pds-amd  Hi. The upsatream aiter PR has merged yesterday. https://github.com/ROCm/aiter/pull/4373 . Could you help review this xDit PR again? Thank you very much!

pds-amd (approved):
@LiuYinfeng01 - thanks for addressing the comments.

LiuYinfeng01:
> @LiuYinfeng01 - thanks for addressing the comments.

Hi @pds-amd ,  thank you again for your review and approval. GitHub previously showed your association as “Contributor,” so it seems the approval was not counted toward the required review. After I re-requested your review, GitHub now recognizes you as a requested reviewer. Would you mind reviewing and approving it once more? Sorry for the inconvenience, and thank you very much!

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -47,18 +47,63 @@
...
+            reset(num_steps)
+
+
+class xFuserWanModel(xFuserModel):
+    """Common lifecycle hooks for Wan runners."""
+
+    def _prepare_inference_run(self, input_args: dict) -> None:
+        _prepare_wan_inference_run(self.pipe, input_args)
+
+    def _setup_hybrid_attn_schedule(self, input_args: dict) -> None:
```
avjves: I think it'd be better to deal with this using config validation. Rather than silently skipping hybrid if using AITER_VSA, it would be better to check it in runtime_state (_check_if_backend_compatible_with_current_configuration, you probably need to pass the info about hybrid schedule to the runtime config object as well). This way there's a proper validation and error handling there and no need to hijack this function.

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -47,18 +47,63 @@
     }
 }
 
+WAN_VSA_ACCURACY_FIRST_DROP_RATES = (0.25, 0.40)
+WAN_VSA_JENGA_REFERENCE_DROP_RATES = (0.75, 0.85)
```
avjves: This is not used anywhere - delete?

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -202,6 +203,9 @@ def __init__(
            pos_embed_seq_len,
         )
         self.attention_kwargs = attention_kwargs
+        self._vsa_denoising_step = -1
```
avjves: Rather than keeping these in the transformer, could you put them in the runtime state? That's where hybrid attention and hybrid gemm state's also live.

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -256,7 +267,41 @@ def forward(
             lora_scale = 1.0
 
 
-        get_runtime_state().increment_step_counter()
+        runtime_state = get_runtime_state()
+        num_steps = (
```
avjves: For clarity, can you put some comments before this block (or move it to a subfunction) that these are all just related to the VSA attention, not the transformer logic.

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -47,18 +47,63 @@
...
+
 
 def _build_attention_kwargs(config: "xFuserArgs") -> dict:
-    """Build the per-model attention_kwargs dict used by the AITER Sparge backends. """
+    """Build shared layout and sparse-attention options for Wan."""
+    drop_rates = config.vsa_drop_rates
+    if config.attention_backend == "AITER_VSA" and not drop_rates:
```
avjves: Compare against the AttentionBackendType Enum, not string

xfuser/model_executor/models/runner_models/wan.py
```diff
@@ -47,18 +47,63 @@
...
+        "use_vsa_static_block_mask": config.use_vsa_static_block_mask,
+        "use_vsa_first_frame_mask": config.use_vsa_first_frame_mask,
+        "vsa_collect_density": config.vsa_collect_density,
     }
 
 
+def _prepare_wan_inference_run(pipe, input_args: dict) -> None:
+    """Publish actual run dimensions and reset per-run VSA state."""
+    num_steps = int(input_args["num_inference_steps"])
+    get_runtime_state().set_video_input_parameters(
```
avjves: I believe you only need the num_steps from this one? 
The other ones are not used anywhere, so no need to set them. It's clearer that way, as some values, such as height/width can still change from the input args (i2v model)

We don't really use the `set_video_input_parameters` anymore, but you can use it if needed :)

xfuser/model_executor/models/transformers/transformer_wan.py
avjves: Also, move the num_steps calculation inside the "if", so non VSA models don't need to reference the runtime state here.

xfuser/core/sparge_attention/sparge.py
```diff
@@ -229,11 +274,21 @@ def setup_sparge(
         if static_mask is not None:
             n_iq = (image_len + img_pad) // block_m
             n_ik = (image_len + img_pad) // block_n
-            padded_static = torch.zeros(
-                (n_iq, n_ik), dtype=static_mask.dtype, device=static_mask.device
-            )
-            padded_static[: static_mask.shape[0], : static_mask.shape[1]] = static_mask
-            static_mask = padded_static
+            if static_mask_cache_key is None:
+                raise RuntimeError("Static block mask cache metadata is missing")
```
avjves: This now changes the behavior of all SPARGE backends. Have you tested the other two sparge backends? Or is the change equivalent to what it was?

avjves (changes requested):
Hey, thanks for the contribution! I left still a few comments. If you could look into them still  then I can approve/merge :)

xfuser/core/sparge_attention/sparge.py
LiuYinfeng01: Thank you for pointing this out. You are right that `setup_sparge` is shared by the existing Sparge backends. I have updated the change so the padded static-mask cache is used only by the VSA-specific sliced-Gilbert path (`use_sliced_gilbert=True`); `AITER_SPARGE` and `AITER_SPARGE_V2` now retain their original allocation behavior. I also added a parameterized regression test covering the existing non-sliced path with sequence reordering both enabled and disabled, verifying that each call still creates a fresh padded mask with identical values. The CPU test suite passed (12 passed, 2 GPU-only tests skipped). I have not run the two AITER Sparge kernels end-to-end in this environment, but the shared setup path they use is now unchanged by construction and covered by the regression test. Thanks again for catching this.

avjves (approved):
Thanks, merging :)

PR #751, opened 2026-08-14
Memory-efficient model loading

pds-amd:
# Memory-efficient model loading

Loading a model eagerly costs one full copy per rank, on the device and on the host. This PR adds two load paths that do not have this resource constraint, leaves the existing paths as the default, and puts the choice behind flags.

- Host memory can stops scaling with rank count!
- Replicated meta-load is worth it even without any sharding for the host memory savings and model load speed up.
- No quality cost. Eager, the naive sharded load and blockwise produce identical outputs.

- **Blockwise FSDP meta fill** builds the transformer on `meta`, then fills and shards one block at a time. It runs behind the existing `--memory_efficient_sharding`, which already sharded block by block but did so from a copy every rank had materialized in host memory first; the new part is reading the checkpoint a block at a time, so that copy never exists.
- **Replicated meta-load** (`--memory_efficient_replicated_load`, new) builds on `meta` and has rank 0 read each block and broadcast it, quantizing before placement. Every rank still ends up with a full copy, so it does not save device memory; using a single checkpoint reader cuts host memory and load time.

Two smaller memory-related additions are included which will be particularly helpful on smaller cards.

- `--use_fp8_text_encoder` extends FP8 to a runner's declared text encoder, opt-in.
- `--enable_group_cpu_offload`, with `--group_offload_low_cpu_mem`, keeps parameters on the host and streams them
to the GPU a group at a time - Faster than sequential offload, slower than model offload.

## Results

meta_load_overview

### Text-encoder FP8

Unlike a change of load path, quantizing does change the sample. Both pairs below are loaded blockwise at eight ranks from the same prompt and seed, and differ only in whether `--use_fp8_gemms --use_fp8_text_encoder` were passed.

FLUX.2-dev at 1024x1024:

| bf16 | FP8 transformer and text encoder |
| --- | --- |
| flux2_blockwise_bf16 |  flux2_blockwise_fp8_te |

Wan2.2-TI2V, 121 frames at 832x1088

bf16

https://github.com/user-attachments/assets/8f359c3a-7179-475c-b2bf-47e8c2c3630b

FP8 transformer and text encoder 

https://github.com/user-attachments/assets/6997d0cf-ff30-402e-a80a-8163699ce2d9

## What changed in the code

The load logic moved out of `base_model` into a new self-contained `loading/` package:

- `contracts.py` holds `LoadDeclaration`, each runner's statement of which components can take which path, and `LoadContract`, the resolved decision. A runner that cannot support a path declares why, and the request is refused before allocation rather than failing partway through a load.
- `meta_load.py`, `shard.py`, `placement.py`, and `checkpoint.py` cover meta construction, the sharded and unsharded placement paths, and block-at-a-time checkpoint reads.
- `backend_selection.py`, `format_backends.py`, `fp8_backends.py`, and `fp8_plan.py` resolve format and backend, including the preflight refusals.
- `text_encoder_plan.py`, `text_encoder_adapter.py`, `transformer_load.py`, `blockwise_ownership.py`, and `quantization_ledger.py` cover the text-encoder and transformer seams, FP8/FP4 ownership within a block, and a single ledger recording what was quantized.

`base_model` keeps two thin seams, `_build_transformer` and `_meta_te_kwargs`, which a runner overrides to opt into a route. AITER FP8 quantization lives in a new `xfuser/model_executor/quant/`.


xfuser/model_executor/models/runner_models/causal_wan.py
```diff
@@ -23,6 +27,14 @@
 
 
 @register_model("CausalWan")
+@LoadDeclaration.declare(
```
avjves: Hmm, would it not be idiomatic to not have these be decorators but define within the model like settings etc?

xfuser/model_executor/models/runner_models/causal_wan.py
```diff
@@ -23,6 +27,14 @@
...
+    loader_adapter=LoaderAdapter.CAUSAL_WAN,
```
avjves: I'm not the biggest fan of this. Now the definition of how a singular model works is defined _outside_ the model object.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -250,10 +273,59 @@ class xFuserModel(abc.ABC):
...
+        self.quantization_ledger = QuantizationLedger()
+
+    def _refresh_load_declaration(self) -> None:
+        """Re-derive FSDP support from instance-customized settings."""
+        declaration = type(self).load_declaration
+        self.load_declaration = LoadDeclaration.for_runner(
+            self.capabilities,
+            meta_transformers=declaration.meta_transformers,
+            replicated=bool(
+                declaration.replicated_meta_transformers
```
avjves: Should this be `declaration.all_replicated_meta_transformers`?

xfuser/model_executor/models/runner_models/loading/shard.py
```diff
@@ -0,0 +1,311 @@
...
+                filter_fn=_target_filter(local_fp4_targets),
+            )
+        if use_fp8_block:
+            adapter = model.blockwise_fp8_backend
+            if adapter is None:
+                raise RuntimeError(
+                    "FP8 block conversion requested without a selected backend"
+                )
+            adapter.convert_block(
+                block,
```
avjves, line 297: Shouldn't this and the block below forward the fp8 suffixes, layers as well?

xfuser/model_executor/models/runner_models/minimax_h3.py
avjves, line 226: Should this also restrict the group offload?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -268,6 +340,16 @@ def _customize_settings(self, config: xFuserArgs) -> None:
     def _update_model_settings(self, config: xFuserArgs) -> None:
         if config.use_fp4_gemms:
             self._apply_fp8_override_cli_from_config(config)
+        te_targets = self.settings.fp8_text_encoder_module_list
+        if config.use_fp8_text_encoder and not te_targets:
+            log(f"--use_fp8_text_encoder has no effect for {type(self).__name__}: it declares no "
```
avjves: Shouldn't this just be gated via the ModelCapabilities?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -304,6 +386,15 @@ def initialize(self, input_args: dict) -> None:
             log("Initializing distributed environment...")
             init_distributed_environment()
 
+        self.load_contract = self._select_preload_contract(
+            world_size=get_world_group().world_size
+        )
+        # Capability/backend mismatches fail here, before _load_model allocates
+        # transformer weights.
+        _ = self.fp8_backend
```
avjves: This seems hacky 😅

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -322,6 +413,119 @@ def initialize(self, input_args: dict) -> None:
                 compile_input_args["prompt"] = compile_input_args["prompt"][: self.config.batch_size]
             self._compile_model(compile_input_args)
 
+    @property
+    def fp8(self):
```
avjves: Hmm, this seems a bit weird.. Why do all models have a fp8 property? Why specifically fp8? Why don't we have plans for fp4, mxfp?

avjves: Also the same about the below other properties. Seems slightly out of place. If needed, should we put properties to the top of the class? Or bottom? A bit of a nit, but properties in the middle of the class just feels wrong 😄

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -322,6 +413,119 @@ def initialize(self, input_args: dict) -> None:
...
+        """Lazy MemoryEfficientLoader bound to this model (meta-init + rank0-broadcast load).
+
+        Must stay cached, unlike ``fp8``: the loader records which transformers it built on meta, so
+        a fresh instance per access would report none of them and silently route every component to
+        the wrong fill path.
+        """
+        from xfuser.model_executor.models.runner_models.loading.meta_load import MemoryEfficientLoader
+        return MemoryEfficientLoader(self)
+
+    def _checkpoint_request(self, subfolder: str | None = None, **kwargs) -> CheckpointRequest:
```
avjves: This too feels like loading related - why is it still in base model, not in loading?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -446,6 +682,10 @@ def _validate_config(self, config: xFuserArgs) -> None:
                 raise ValueError(f"Model {self.settings.model_name} does not support distilled_transformer_path or distilled_transformer_2_path params.")
 
 
+    # torch.compile modes that run the graph under CUDA Graphs, whose outputs live in a fixed buffer
+    # pool and are therefore only valid until the next replay.
```
avjves: Odd place for this? It's used in a single function below. Why hoist it here?

avjves: And if cleaner, maybe it too should be somewhere higher up in code

xfuser/model_executor/models/runner_models/causal_wan.py
avjves: If this is special case, it'd be better if this was another capability / requirement / restriction like instead of specific adapter. Something `LoaderRequirement.SUPPORTS_BLOCKWISE` (I don't remember what it was actually for anymore 😄 )

avjves (commented):
Thanks for this!
This is great, both feature and refactorwise. I love that the quantization and sharding stuff are leaving base model 😄 This is quite a large PR again though, so I'm having a hard time reviewing it 😓 If possible, please smaller PRs in the future 🫣 

I left a few comments again, nothing really major though :)

As this changes _a lot_ of things, I think we should run the full CI + supported models not in it and compare to previous known numbers.

xfuser/model_executor/models/runner_models/causal_wan.py
pds-amd: I went back and forth on this and am happy to change it!

The decorator’s initial motivation was that it runs after class creation, so it can derive the declaration from both ModelCapabilities and ModelSettings and felt like the least invasive way to add this as a self-contained feature.

Would you prefer a separate class-level load_spec, defined inside the model class and resolved against the final instance settings during initialization?

avjves: Yeah in isolation there's nothing wrong it - seems like a nice way to do it. It's just that there's already the ModelSettings, ModelCapabilities, DefaultInputValues. So maybe it makes sense to move this next to these?

avjves: From what I gather the declaration was re-created in init? Does it pose a problem if we move it to be next to those other settings. If problematic, we don't have to move :)

pds-amd: I'll do some experiments today to see if I can make this more ergonomic. I agree with this and other comments you have made on this PR. I'll ping you later with any proposed changes.

pds-amd: Changed this to a class-level `load_spec` beside the runner settings/capabilities. It now stores only static intent and is resolved against the instance settings after `_customize_settings`, so the model definition stays inside the class without duplicating or freezing runtime settings.

xfuser/model_executor/models/runner_models/minimax_h3.py
pds-amd: Yes. MiniMax-H3 text-encoder TP now rejects group offload as well as model and sequential offload, with parametrized coverage for all three modes.

xfuser/model_executor/models/runner_models/causal_wan.py
pds-amd: Agreed and changed. Every registered runner now owns an explicit class-level `load_spec`. I also replaced the named loader-adapter special cases with independent `STANDARD_COLLECTIVES` and `LOCAL_BLOCKWISE` capabilities; the custom checkpoint reason remains in the model-local refusal text.

xfuser/model_executor/models/runner_models/loading/shard.py
pds-amd: The precision prefix/suffix overrides describe the FP8 remainder owned by an FP4 conversion, so forwarding them to pure FP8 or INT8 adapters would be the wrong contract. I kept them on the FP4 call. I did find and fix the adjacent issue: blockwise pure-FP8 conversion now honors `fp8_gemm_include_suffixes`, matching eager placement, and native streaming falls back when it cannot represent that suffix restriction.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Changed this to the standard capability path. `ModelCapabilities` now has `use_fp8_text_encoder`, supported runners declare it alongside their targets, and other runners reject the flag during validation. There is also a registry invariant keeping the capability and target declaration in sync.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: This reconstruction is gone now. The new static `LoadSpec` preserves the explicit `replicated` intent directly, and `ModelLoader` resolves the component tuples once against the final instance settings.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Agreed. `initialize` now makes one explicit `self.loader.preflight(...)` call. Backend probing/validation is owned by that method and only resolves the adapters required by the selected contract; the property-forcing lines are gone from the runner.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Changed this to a backend-neutral `QuantizationPlan` owned by the run's single `ModelLoader`. It now centralizes FP8, FP4 and INT8 target lookup, including the optional text-encoder coverage. The format-specific properties have been removed from every model.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Moved it into `ModelLoader`. Default checkpoint request construction, offline behavior and request merging are now loading-owned; Hunyuan only declares class-level request defaults for its pinned revision.

xfuser/model_executor/models/runner_models/base_model.py
pds-amd: Moved `CUDAGRAPH_COMPILE_MODES` to the class constants at the top of `xFuserModel`.

avjves:
A couple of new merge conflicts here as well due to another PR being merged - can you add the load declaration to LTX2.5 as well? :)

avjves (approved):

PR #752, opened 2026-08-17
integrate new AITER mha v4 fmha entrypoint to xDiT

jcaraban:
## Motivation

Expose the low-precision MHA v4 kernels added by https://github.com/ROCm/aiter/pull/4627 through xDiT.

## Technical Details

- Add I8FP8, MXFP4, MXFP6, F4F4, and F6F4 attention backends.
- Delegate preprocessing and kernel dispatch to AITER's public `mha_v4` API.
- Support unequal query and KV sequence lengths.
- Validate availability through the public AITER API.
- Reject unsupported causal attention explicitly.
- Document the new gfx950, head-dimension-128 backends.

## Testing

- Added eager correctness checks against SDPA.
- Added `torch.compile(fullgraph=True)` coverage.
- Covered unequal sequence lengths, cross-attention, partial tiles, and causal rejection.
- All 20 focused tests pass on gfx950.


xfuser/core/distributed/attention_backend.py
```diff
@@ -373,6 +373,32 @@ def _build_aiter_mla_metadata(batch_size, q_seq_len, kv_seq_len, num_heads, num_
...
+        _AITER_MHA_V4_AVAILABLE = (
+            torch.cuda.is_available()
+            and "gfx950" in torch.cuda.get_device_properties(0).gcnArchName
+        )
+    except ImportError:
+        _AITER_MHA_V4_AVAILABLE = False
+        pass # Error is raised in runtime_state.py when an MHA v4 backend is selected.
+
+    try:
+        from aiter.ops.mha_v4 import (
```
avjves, line 396: Why are these separate from the above check?

avjves (approved):

xfuser/core/distributed/attention_backend.py
jcaraban: because `mha_v4_mxfp8` will fail to import until https://github.com/ROCm/aiter/pull/4967 merges, so if it was with the above block it would make the other MHA v4 backends fail, while those import fine in AITER main

PR #753, opened 2026-08-18
FP8 Ulysses All-to-All Communication

tjkemp:
Quantizes Ulysses all-to-all Q/K/V/O traffic to FP8 with per-layer calibrated scales, reducing communication volume by ~2x vs BF16.

The whole feature is encapsulated in `xfuser/core/distributed/fp8_comms.py`. The rest of the codebase (runtime_state, base_model, usp, transformer_wan) only makes thin calls into it.

Supported attention backends: `AITER_FP8`. Enabled models: `Wan2.2-I2V`, `Wan2.2-T2V`, `Wan2.2-TI2V`.

## Usage

Add `--use_fp8_comms` to an existing Wan2.2 run that already uses Ulysses.

Two optional args:

- `--fp8_comms_scale FLOAT` : use a fixed scale for every layer and skip   calibration entirely.
- `--fp8_comms_safety_factor FLOAT` : headroom applied to the calibrated scale (default 0.85). Lower value maps the calibrated amax further below the FP8 max, trading a little dynamic range for fewer overflows.

## How it works

The feature has three parts: a one-time calibration, a quantized all-to-all on each step, and an attention call that consumes the already-quantized tensors.

### 1. Calibration (once per run, before the timed steps)

FP8 has a small dynamic range, so each tensor needs a scale that maps its values into that range without clipping. Rather than compute a scale on every step, the feature measures it once:

- One throwaway BF16 forward pass runs before inference.
- During that pass, each self-attention layer records the maximum absolute value (amax) of the Q, K, V and output tensors it will later quantize.
- The per-layer amaxes are all-reduced with a max across the Ulysses ranks, so every rank agrees on the same scale for a given layer.
- Each per-layer scale is computed as `amax / (FP8_MAX * safety_factor)` and frozen into small buffers on the attention module.

Calibration finishes before `torch.compile` captures the graph, so the scales are constants in the compiled region. Nothing in the calibrated path runs inside the compiled graph, so there are no graph breaks on the hot path.

If `--fp8_comms_scale` is given, calibration is skipped and that fixed value is used for every layer.

### 2. Quantized all-to-all (every step)

Inside Ulysses, for each self-attention layer:

- Q and K are rotated by the existing Hadamard rotation (already merged) before   quantization. The rotation preserves the Q times K product but spreads outliers,  which lowers FP8 quantization error at no change to the attention result.
- Q, K and V are quantized to FP8 using the frozen per-layer scales.
- The all-to-all exchanges the FP8 tensors.
- After attention, the output is quantized to FP8 for the output all-to-all and   dequantized back to BF16 on the receiving side.

### 3. Attention on pre-quantized inputs

Because the tensors arrive already in FP8, the attention kernel does not quantize again. The AITER_FP8 path takes a `pre_quantized` flag: it feeds the FP8 Q/K/V and their descale factors straight into the dense FP8 attention kernel
(`flash_attn_fp8_pertensor_func`) and returns. One quantization for both transport and compute, no dequantize-and-requantize in between.

## Performance

Wan2.2, 8x GPU Ulysses, `AITER_FP8` hybrid backend, MI355X:

| Task  | Speedup |
|---|---|
| I2V | +2.88% |
| T2V | +3.13% |

## Example

```bash
--model Wan2.2-T2V --ulysses_degree 8 --use_hybrid_attn_schedule --hybrid_attn_low_precision_backend aiter_fp8 --hybrid_attn_high_precision_backend aiter --num_hybrid_attn_high_precision_steps 3 --use_fp8_gemms --use_fp8_comms
```

### DIFF
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -11,6 +11,7 @@ import torch.distributed
 from xfuser.logger import init_logger
 from xfuser.core.distributed import init_distributed_environment
 from xfuser.config.config import (
+    DEFAULT_FP8_COMMS_SAFETY_FACTOR,
     EngineConfig,
     FastAttnConfig,
     ParallelConfig,
@@ -152,6 +153,9 @@ class xFuserArgs:
     use_fp4_gemms: bool = False
     fp8_precision_override_prefix_patterns: Optional[str] = None
     fp8_precision_override_suffix_patterns: Optional[str] = None
+    use_fp8_comms: bool = False
+    fp8_comms_scale: Optional[float] = None
+    fp8_comms_safety_factor: float = DEFAULT_FP8_COMMS_SAFETY_FACTOR
     # Model runner specific
     num_iterations: int = 1
     profile: bool = False
@@ -517,6 +521,25 @@ class xFuserArgs:
             action="store_true",
             help="Quantize the transformer linear layers (selected models only).",
         )
+        runtime_group.add_argument(
+            "--use_fp8_comms",
+            action="store_true",
+            help="Quantize Ulysses all-to-all communication to FP8.",
+        )
+        runtime_group.add_argument(
+            "--fp8_comms_scale",
+            type=float,
+            default=None,
+            help="Override the model-specific FP8 communication scale.",
+        )
+        runtime_group.add_argument(
+            "--fp8_comms_safety_factor",
+            type=float,
+            default=DEFAULT_FP8_COMMS_SAFETY_FACTOR,
+            help="Safety factor for the calibrated FP8 comms scale: scale = amax / "
+                 "(FP8_MAX * safety_factor). LOWER it (e.g. 0.4) to enlarge the scale and "
+                 "leave more headroom before fp8 saturation. Default 0.85.",
+        )
 
         # DiTFastAttn arguments
         fast_attn_group = parser.add_argument_group("DiTFastAttn Options")
@@ -814,7 +837,25 @@ class xFuserArgs:
             default=None,
             help="Comma-delimited FQN suffix patterns to keep in FP8 during FP4 GEMMs.",
         )
-
+        parser.add_argument(
+            "--use_fp8_comms",
+            action="store_true",
+            help="Quantize Ulysses all-to-all communication to FP8.",
+        )
+        parser.add_argument(
+            "--fp8_comms_scale",
+            type=float,
+            default=None,
+            help="Override the model-specific FP8 communication scale.",
+        )
+        parser.add_argument(
+            "--fp8_comms_safety_factor",
+            type=float,
+            default=DEFAULT_FP8_COMMS_SAFETY_FACTOR,
+            help="Safety factor for the calibrated FP8 comms scale: scale = amax / "
+                 "(FP8_MAX * safety_factor). LOWER it (e.g. 0.4) to enlarge the scale and "
+                 "leave more headroom before fp8 saturation. Default 0.85.",
+        )
         parser.add_argument(
             "--num_iterations",
             type=int,
@@ -1201,6 +1242,9 @@ class xFuserArgs:
             use_vsa_static_block_mask=self.use_vsa_static_block_mask,
             use_vsa_first_frame_mask=self.use_vsa_first_frame_mask,
             vsa_collect_density=self.vsa_collect_density,
+            use_fp8_comms=self.use_fp8_comms,
+            fp8_comms_scale=self.fp8_comms_scale,
+            fp8_comms_safety_factor=self.fp8_comms_safety_factor,
         )
 
         parallel_config = ParallelConfig(
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -144,12 +144,11 @@ def _build_hadamard_matrix(block_r, dtype=torch.bfloat16, allow_sylvester_fallba
 
 def _replicate_hadamard_per_device(hadamard):
     """Replicate a single Hadamard matrix on each available device, keyed by
-    torch.device (all GPUs if CUDA is available, else CPU). A None matrix maps
+    torch.device (CPU plus all GPUs when CUDA is available). A None matrix maps
     to None on every device."""
+    devices = [torch.device("cpu")]
     if torch.cuda.is_available():
-        devices = [torch.device(f"cuda:{i}") for i in range(torch.cuda.device_count())]
-    else:
-        devices = [torch.device("cpu")]
+        devices += [torch.device(f"cuda:{i}") for i in range(torch.cuda.device_count())]
     return {
         device: (hadamard.to(device) if hadamard is not None else None)
         for device in devices
@@ -389,8 +388,11 @@ def _build_aiter_mla_metadata(batch_size, q_seq_len, kv_seq_len, num_heads, num_
 
 aten = torch.ops.aten
 env_info = PACKAGES_CHECKER.get_packages_info()
+AITER_FP8_DTYPE = torch.float8_e4m3fn  # fallback; fp8 comms requires aiter
+FP8_HADAMARD_MATRIX = _aiter_hadamard_matrix(128)
 if env_info["has_aiter"]:
     import aiter
+    AITER_FP8_DTYPE = aiter.dtypes.fp8
     from aiter import flash_attn_func as flash_attn_func_aiter
     from aiter import flash_attn_varlen_func as flash_attn_varlen_func_aiter
     try:
@@ -689,6 +691,13 @@ def _mha_v4_sparge_tile():
     return {"BLOCK_M": 256, "BLOCK_N": _AITER_MHA_V4.kv_tile}
 
 
+_FP8_INPUT_DTYPES = (torch.float8_e4m3fn, torch.float8_e4m3fnuz)
+
+SUPPORTS_PRE_QUANTIZATION_BACKENDS = {
+    AttentionBackendType.AITER_FP8,
+}
+
+
 def register_attention_function(backend_type):
     """
     Decorator to register attention functions with their corresponding backend type.
@@ -951,6 +960,24 @@ def _fp8_hadamard_rotate(x: torch.Tensor, R: torch.Tensor) -> torch.Tensor:
     return torch.matmul(x.unflatten(-1, (d // block_r, block_r)), R).flatten(-2)
 
 
+def rotate_qk_for_fp8_comms(query, key, backend):
+    """Rotate Q,K exactly as the fp8-comms path does before quantizing them.
+
+    Returns the inputs untouched for backends whose fp8-comms path does not rotate,
+    so callers can apply this unconditionally. Single definition shared by the
+    quantization site in USP and the calibration site in the attention processor:
+    both must measure and quantize the same distribution, otherwise the frozen
+    per-layer scale describes a tensor that is never quantized.
+    """
+    if backend != AttentionBackendType.AITER_FP8:
+        return query, key
+    R = _get_fp8_hadamard_matrix(query.shape[-1], query.device)
+    return (
+        _fp8_hadamard_rotate(query, R).contiguous(),
+        _fp8_hadamard_rotate(key, R).contiguous(),
+    )
+
+
 def _quantize_aiter_fp8_inputs(query, key, value):
     quant_dtype = aiter.dtypes.fp8
     dtype_max = torch.finfo(quant_dtype).max
@@ -1102,9 +1129,28 @@ def _aiter_fp8_attn_call(query, key, value, dropout_p, is_causal, attention_kwar
     then calls attention through AITER
     """
     _validate_aiter_low_precision_dropout(dropout_p)
+    attention_kwargs = attention_kwargs or {}
+    pre_quantized = attention_kwargs.get("pre_quantized", False)
+
     query = torch.permute(query, [0, 2, 1, 3]).contiguous()
     key = torch.permute(key, [0, 2, 1, 3]).contiguous()
     value = torch.permute(value, [0, 2, 1, 3]).contiguous()
+
+    if pre_quantized:
+        # Q/K/V arrive already FP8 from fp8 comms (quantized before the Ulysses
+        # all-to-all).
+        output = aiter.flash_attn_fp8_pertensor_func(
+            query,
+            key,
+            value,
+            causal=is_causal,
+            softmax_scale=query.shape[-1] ** -0.5,
+            q_descale=attention_kwargs["q_descale"],
+            k_descale=attention_kwargs["k_descale"],
+            v_descale=attention_kwargs["v_descale"],
+        )
+        output = torch.permute(output, [0, 2, 1, 3])
+        return output, None
 
     packed = _varlen_pack_keys(query, key, value, attention_kwargs)
     use_mha_v4 = packed is None and _use_aiter_mha_v4_fp8(query, is_causal)
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -691,8 +691,6 @@ def _mha_v4_sparge_tile():
     return {"BLOCK_M": 256, "BLOCK_N": _AITER_MHA_V4.kv_tile}
 
 
-_FP8_INPUT_DTYPES = (torch.float8_e4m3fn, torch.float8_e4m3fnuz)
-
 SUPPORTS_PRE_QUANTIZATION_BACKENDS = {
     AttentionBackendType.AITER_FP8,
 }
--- a/xfuser/core/distributed/attention_backend.py
+++ b/xfuser/core/distributed/attention_backend.py
@@ -143,10 +143,9 @@ def _build_hadamard_matrix(block_r, dtype=torch.bfloat16, allow_sylvester_fallba
 
 
 def _replicate_hadamard_per_device(hadamard):
-    """Replicate a single Hadamard matrix on each available device, keyed by
-    torch.device (CPU plus all GPUs when CUDA is available). A None matrix maps
-    to None on every device."""
-    devices = [torch.device("cpu")]
+    """Replicate a single Hadamard matrix on each visible GPU, keyed by
+    torch.device. A None matrix maps to None on every device."""
+    devices = []
     if torch.cuda.is_available():
         devices += [torch.device(f"cuda:{i}") for i in range(torch.cuda.device_count())]
     return {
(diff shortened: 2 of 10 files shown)


xfuser/config/args.py
```diff
@@ -729,7 +752,26 @@ def add_runner_args(parser: FlexibleArgumentParser):
...
+        )
+        parser.add_argument(
+            "--fp8_comms_safety_factor",
+            type=float,
+            default=0.85,
+            help="Safety factor for the calibrated FP8 comms scale: scale = amax / "
+                 "(FP8_MAX * safety_factor). LOWER it (e.g. 0.4) to enlarge the scale and "
+                 "leave more headroom before fp8 saturation, removing static-calibration "
+                 "clipping at zero runtime cost (over-scaling is lossless for fp8). Default 0.85.",
+        )
```
lauri9, line 858: These were already added to `runtime_group`, so this is a duplicate?

xfuser/core/distributed/runtime_state.py
```diff
@@ -48,6 +54,8 @@
 logger = init_logger(__name__)
 
 env_info = PACKAGES_CHECKER.get_packages_info()
+_FP8_COMMS_SAFETY_FACTOR = 0.85
```
lauri9: I don't like how this default is set in many places:
```python
# config.py / args.py
fp8_comms_safety_factor: float = 0.85
# runtime_state.py
_FP8_COMMS_SAFETY_FACTOR = 0.85
```

Can this be specified canonically in one place?

xfuser/core/distributed/runtime_state.py
```diff
@@ -57,9 +65,137 @@ def set_random_seed(seed: int):
...
+            block.attn1.fp8_q_scale.copy_(q_scales[i : i + 1])
+            block.attn1.fp8_k_scale.copy_(k_scales[i : i + 1])
+            block.attn1.fp8_v_scale.copy_(v_scales[i : i + 1])
+            block.attn1.fp8_o_scale.copy_(o_scales[i : i + 1])
+
+    def to_device_(self, device: torch.device):
+        for model_state in self._models.values():
+            model_state.to_device_(device)
+
+
```
lauri9: Could these classes and helpers live in a separate file, to clearly distinguish FP8 feature?

xfuser/core/distributed/runtime_state.py
```diff
@@ -129,6 +266,75 @@ def set_attention_backend(self, attention_backend: str | AttentionBackendType):
...
+        scale = config.runtime_config.fp8_comms_scale
+        safety_factor = config.runtime_config.fp8_comms_safety_factor
+        if scale is not None:
+            logger.warning(f"FP8 communication enabled with fixed scale {scale}.")
+        else:
+            logger.warning(
+                "FP8 communication enabled with static per-layer scaling "
+                f"(calibrated once before inference; safety_factor={safety_factor})."
+            )
+        self.fp8_comms = Fp8CommsState(fixed_scale=scale, safety_factor=safety_factor)
```
lauri9: This method is only initializing `self.fp8_comms`. Can't we handle this logic in the constructor of `Fp8CommsState`, not as a method in this class? Feels out of place.

xfuser/core/distributed/runtime_state.py
```diff
@@ -129,6 +266,75 @@ def set_attention_backend(self, attention_backend: str | AttentionBackendType):
...
+            print(
+                f"[fp8_comms] {model.__class__.__name__} per-layer scales synced: "
+                f"q=[{q_scales.min().item():.6f}, {q_scales.max().item():.6f}] "
+                f"k=[{k_scales.min().item():.6f}, {k_scales.max().item():.6f}] "
+                f"v=[{v_scales.min().item():.6f}, {v_scales.max().item():.6f}] "
+                f"o=[{o_scales.min().item():.6f}, {o_scales.max().item():.6f}] "
+                f"(amax q={maxes[0].max().item():.4f} k={maxes[1].max().item():.4f} "
+                f"v={maxes[2].max().item():.4f} o={maxes[3].max().item():.4f})"
+            )
+
```
lauri9: Likewise - does this have to be a method of this class?

xfuser/core/distributed/attention_backend.py
```diff
@@ -829,6 +842,24 @@ def _fp8_hadamard_rotate(x: torch.Tensor, R: torch.Tensor) -> torch.Tensor:
     return torch.matmul(x.unflatten(-1, (d // block_r, block_r)), R).flatten(-2)
 
 
+def rotate_qk_for_fp8_comms(query, key, backend):
```
lauri9, line 963: Should `_aiter_fp8_attn_call` in this file should use this helper? Now, it is still inlining the rotation.

This helper name claims it is only for FP8 comms purpose - but is it really limited to that use case only? Should the naming also change?

xfuser/core/distributed/attention_backend.py
```diff
@@ -829,6 +842,24 @@ def _fp8_hadamard_rotate(x: torch.Tensor, R: torch.Tensor) -> torch.Tensor:
...
+    """Rotate Q,K exactly as the fp8-comms path does before quantizing them.
+
+    Returns the inputs untouched for backends whose fp8-comms path does not rotate,
+    so callers can apply this unconditionally. Single definition shared by the
+    quantization site in USP and the calibration site in the attention processor:
+    both must measure and quantize the same distribution, otherwise the frozen
+    per-layer scale describes a tensor that is never quantized.
+    """
+    if backend != AttentionBackendType.AITER_FP8:
+        return query, key
```
lauri9, line 973: If it is mandatory to have `AITER_FP8` backend, why not raise here instead of silently returning non-rotated tensors?

xfuser/model_executor/layers/usp.py
```diff
@@ -108,6 +119,56 @@ def _ft_c_input_all_to_all(x):
...
+
+
+def _fp8_comms_output_all_to_all(out: torch.Tensor, o_scale_t: torch.Tensor) -> torch.Tensor:
+    """Quantize attention output to FP8, run output all-to-all, dequantize back."""
+    restore_dtype = out.dtype if out.dtype not in _FP8_DTYPES else torch.bfloat16
+    if out.dtype not in _FP8_DTYPES:
+        out_fp8, out_descale = _per_tensor_quant(out, o_scale_t)
+    else:
+        out_fp8, out_descale = out, o_scale_t
+    return (_ft_c_output_all_to_all(out_fp8).float() * out_descale).to(restore_dtype)
```
lauri9: These are also candidates for functions or methods that could live in the separate FP8 comms feature file.

xfuser/model_executor/layers/usp.py
```diff
@@ -302,8 +368,25 @@ def USP(
 
         }
 
+    qkv_scales = None
```
lauri9: Why init `qkv_scales` here?

xfuser/core/distributed/runtime_state.py
```diff
@@ -129,6 +266,75 @@ def set_attention_backend(self, attention_backend: str | AttentionBackendType):
...
+        fp8_comms._scatter_scales_to_model(model, scales[0], scales[1], scales[2], scales[3])
+        model_state.q_running_max.zero_()
+        model_state.k_running_max.zero_()
+        model_state.v_running_max.zero_()
+        model_state.o_running_max.zero_()
+        model_state.synced = True
+        fp8_comms.calibrated_model_ids.add(id(model))
+        if dist.get_rank() == 0:
+            q_scales, k_scales, v_scales, o_scales = scales[0], scales[1], scales[2], scales[3]
+            print(
```
lauri9: Can we use logging module instead of print

xfuser/model_executor/layers/usp.py
```diff
@@ -350,7 +433,19 @@ def USP(
                             is_causal=is_causal,
                             joint_attn_kwargs=joint_attn_kwargs,
                             attention_kwargs=attention_kwargs)
-        out = _ft_c_output_all_to_all(out)
+        if use_fp8_comms:
+            if _FP8_LOG_SCALES and qkv_amaxes is not None:
+                out_amax = out.abs().amax().item()
+                rank = dist.get_rank()
+                q_amax, k_amax, v_amax = qkv_amaxes
+                print(f"[fp8_scales rank{rank}] q_amax={q_amax:.4f} k_amax={k_amax:.4f} v_amax={v_amax:.4f} out_amax={out_amax:.4f}")
```
lauri9: Can we use logging module instead of print?

xfuser/model_executor/layers/usp.py
```diff
@@ -256,9 +317,14 @@ def USP(
...
         attn_layer=None,
         combine_qkv_a2a: bool | None = None,
+        use_fp8_comms: bool = False,
         backend=None,
         attention_kwargs: dict | None = None,
         head_balance_layer=None,
+        fp8_q_scale: torch.Tensor | None = None,
+        fp8_k_scale: torch.Tensor | None = None,
+        fp8_v_scale: torch.Tensor | None = None,
+        fp8_o_scale: torch.Tensor | None = None,
```
lauri9: Instead of changing the function signature to include `use_fp8_comms` and `fp8_q_scale`, etc., can we instead pass one optional `FP8Comms` object or something, that would encapsulate all the required info?

I think changing the signature this much can be a bit confusing, for an optional feature

xfuser/model_executor/layers/usp.py
```diff
@@ -302,8 +368,25 @@ def USP(
...
+            query, key = rotate_qk_for_fp8_comms(query, key, fp8_comms_backend)
+            if fp8_q_scale is None or fp8_k_scale is None or fp8_v_scale is None:
+                raise RuntimeError(
+                    "FP8 comms requires per-layer scale buffers (fp8_q_scale, fp8_k_scale, fp8_v_scale)."
+                )
+            query, key, value, attn_kwargs_update, qkv_scales, qkv_amaxes = _fp8_comms_input_all_to_all(
+                query, key, value, fp8_q_scale, fp8_k_scale, fp8_v_scale,
+            )
+            attention_kwargs = (attention_kwargs or {}) | attn_kwargs_update
+        elif combine_qkv_a2a and query.shape == key.shape == value.shape:
```
lauri9, line 329: I'd like to encapsulate this more or hoist the logic to a separate function to make this easier to understand.

xfuser/model_executor/layers/usp.py
```diff
@@ -350,7 +433,19 @@ def USP(
...
+            if _FP8_LOG_SCALES and qkv_amaxes is not None:
+                out_amax = out.abs().amax().item()
+                rank = dist.get_rank()
+                q_amax, k_amax, v_amax = qkv_amaxes
+                print(f"[fp8_scales rank{rank}] q_amax={q_amax:.4f} k_amax={k_amax:.4f} v_amax={v_amax:.4f} out_amax={out_amax:.4f}")
+            if fp8_o_scale is None:
+                raise RuntimeError(
+                    "FP8 comms requires per-layer scale buffers fp8_o_scale"
+                )
+            out = _fp8_comms_output_all_to_all(out, fp8_o_scale)
```
lauri9: Could this logic also be hoisted elsewhere?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -347,6 +354,79 @@ def _decoding_vaes(self) -> List:
...
+                    )
+        if not effective_backends & SUPPORTS_PRE_QUANTIZATION_BACKENDS:
+            raise ValueError(
+                f"--use_fp8_comms requires an attention backend that supports pre-quantization "
+                f"({', '.join(b.name for b in SUPPORTS_PRE_QUANTIZATION_BACKENDS)}). "
+                f"Set --attention_backend, --hybrid_attn_schedule, or "
+                f"--hybrid_attn_low_precision_backend / --hybrid_attn_high_precision_backend "
+                f"so at least one scheduled backend supports pre-quantization."
+            )
+
```
lauri9: Regarding these changes. It feels like these are responsibility of a FP8 feature class? Can we invert the dependency here that base model is not responsible for managing all of this.

xfuser/model_executor/models/runner_models/wan.py
lauri9, line 1: Why `capabilities = replace(xFuserWan21I2VModel.capabilities, use_fp8_comms=True)`? Do we need to limit the applicability somehow? This feels a bit ugly and makes it hard to reason about model capabilities

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -122,6 +127,54 @@ def apply_rotary_emb(
...
+                    calib_query, calib_key = rotate_qk_for_fp8_comms(
+                        query, key, runtime_state.attention_backend
+                    )
+                    runtime_state.fp8_comms.update_running_max(
+                        fp8_owner,
+                        attn.fp8_comms_layer_idx,
+                        calib_query,
+                        calib_key,
+                        value,
+                    )
```
lauri9: Could the logic for updates also live in the separate class, I don't see why it has to be part of transformer

xfuser/model_executor/models/transformers/transformer_wan.py
```diff
@@ -131,27 +184,41 @@ def apply_rotary_emb(
...
+        # update running max for dynamic scale calibration -- pure in-place tensor ops, no graph break
+        if not self.is_cross_attention and runtime_state.fp8_comms is not None:
+            fp8_owner = getattr(attn, "fp8_comms_owner", None)
+            if fp8_owner is not None and hasattr(attn, "fp8_comms_layer_idx"):
+                runtime_state.fp8_comms.update_o_running_max(
+                    fp8_owner,
+                    attn.fp8_comms_layer_idx,
+                    hidden_states,
+                )
+
```
lauri9: Update logic could be in fp8 class?

lauri9 (changes requested):
This is a good feature, but I think the implementation is too heavily coupled into the existing attention/USP/base model implementations. I would like to see the FP8 comms part and its logic hoisted into a separate file to encapsulate it better, and minimize its coupling to the other classes.

There are also many minor issues that should be addressed before merging.

xfuser/config/args.py
tjkemp: The duplication isn't something this PR introduced. Every argument in this codebase is already declared twice. `--use_fp8_gemms` for instance appears on lines 426 and 668.

lauri9 (changes requested):
Thanks for reworking the feature, it's much cleaner separated into its own file. It's working as advertised and I cannot find any bugs.

There are a few small items that still need to be updated. Otherwise it looks good and does not seem to impact behavior when the feature is disabled.

# Failing unit test
PR regresses one of the VAE lifecycle tests still:
```
FAILED tests/core/test_vae_lifecycle.py::test_initialize_sets_up_every_parallel_vae_before_enabling_options
E   AttributeError: 'Runner' object has no attribute '_run_timed_pipe'
xfuser/model_executor/models/runner_models/base_model.py:357: AttributeError
```

It's best to guard the call to `setup_fp8_comms` instead, such as:
```python
fp8_comms = get_runtime_state().fp8_comms if runtime_state_is_initialized() else None
if fp8_comms is not None:
    setup_fp8_comms(
        fp8_comms, self.pipe, input_args,
        run_pipe_fn=self._run_timed_pipe,
        split_prompts_fn=self._split_prompts_for_dp,
        batch_size=self.config.batch_size,
    )
```

# `safety_factor > 1` should not be accepted

We have `scale = amax / (FP8_MAX * safety_factor)`, so a `safety_factor` above 1 shrinks the scale. This is not helpful, so we can guard against it:
```python
if not 0 < safety_factor <= 1:
    raise ValueError(
        f"fp8_comms_safety_factor must be in (0, 1], got {safety_factor}."
    )
```

# `HADAMARD_MATRIX` guards
`FP8_HADAMARD_MATRIX =
_aiter_hadamard_matrix(128)` sits at `attention_backend.py:392`, above the
`if env_info["has_aiter"]:` gate, and `_replicate_hadamard_per_device` does `hadamard.to(device)`
for every visible GPU. On base, both Hadamard matrices were built inside the gate, so a non-AITER
host built none. A non-AITER host now touches every visible GPU at import. Guard the build with
`env_info["has_aiter"]`, or make it lazy.

# `HADAMARD_MATRIX` CPU key
The PR introduces a `"cpu"` key for hadamard matrix, but is this only needed for unit test? To limit the changes, please mock this in unit tests instead rather than changing production code.

lauri9 (approved):
👍

xfuser/core/distributed/fp8_comms.py
```diff
@@ -0,0 +1,551 @@
...
+from xfuser.config.config import DEFAULT_FP8_COMMS_SAFETY_FACTOR
+
+if torch.cuda.is_available() or envs._is_npu():
+    from yunchang.globals import PROCESS_GROUP
+else:
+    PROCESS_GROUP = None
+
+logger = init_logger(__name__)
+
+_FP8_LOG_SCALES = bool(os.environ.get("XFUSER_FP8_LOG_SCALES"))
```
avjves: nit: XFUSER_FP8_LOG_SCALES=0 would be True here

xfuser/core/distributed/fp8_comms.py
```diff
@@ -0,0 +1,551 @@
...
+            batch_size=batch_size,
+        )
+
+
+def validate_fp8_comms_config(config, capabilities, settings) -> None:
+    """Raise if --use_fp8_comms is requested but unsupported by the model/config; no-op if off."""
+    if not config.use_fp8_comms:
+        return
+    from xfuser.core.distributed.attention_backend import (
+        AITER_FP8_HAS_DESCALE,
```
avjves: AITER_FP8_HAS_DESCALE is AITER specific - will always crash on NV. Should be some other error than import error 🤔

avjves (approved):
Thanks! Left a couple of comments.

I looked at this from code perspective and there's two things that came to mind:
1. fp8_comms now finds itself as a separate parameter in many places, like usp call signature. What about fp6_comms? Cleaner would be to have a generic comms object, which is then setup to be fp8. Might be there's no plan to go below fp8, so OK for now. 
2. The attention now calls specific fp8 impl when using this. Does it only work with it? So the other FP8 implementations do not? The attention backend is quite... verbose currently and _will_ receive a re-write sometime soon, so added complexity doesn't hurt much 😄 

If you think neither require effort right now, then after looking at the comments we can merge :)

tjkemp:
A comms object sounds good! I'll leave that for the future once there are new parameters. This work targets the specific FP8 implementation. The attention kernel must 1) accept FP8 Q/K/V, 2) accept per-tensor descales, and 3) not do its own rotation or quantization.

PR #757, opened 2026-08-21
Fix broken "Performance" anchor in Table of Contents

GlowLED:
Since commit `c539c6e` (PR #475) merged the standalone "Performance Reports"
section into the "Supported DiTs" table, the Table of Contents entry
`[📈 Performance](#perf)` pointed to a non-existent `#perf` anchor.

Point it at `#support-dits`, where the performance report links now live.


avjves (approved):
Thanks for the docs fix, they easily end up being outdated 😁

PR #758, opened 2026-08-24
MHAv4 sparse attention integration

nsakkine:
## Summary

Add eight `aiter_*_sparge` attention backends that reuse the existing Sparge block mask, built at the tile geometry of the active MHA v4 sparse row, and launch the matching AITER MHA v4 sorted-sparse kernel. Dense siblings (`aiter_fp8`, `aiter_mxfp4`, and the rest) stay as they are. Triton Sparge and Flex Sparge are unchanged.

This is the xDiT half of gfx950/gfx942 block-sparse FMHA. It needs an AITER build whose `mha_v4` accepts `block_mask` and ships the matching `fwd_hd128_*_sparse.co`.

Arch coverage:

- gfx950: all eight recipes, 256x128 tiles
- gfx942: `aiter_fp8_sparge` and `aiter_i8fp8_sparge` only, 256x64 tiles

## What changed

- `xfuser/core/distributed/attention_backend.py`: eight backends plus shared `_aiter_mha_v4_sparge_call`. Import-time probes record whether sparse is available, whether the GPU is gfx942, whether `mha_v4` / `mha_v4_mxfp8` accept `block_mask`, and the sparse KV tile (from `aiter.ops.mha_v4.mha_v4_kv_tile()` when present, else 64 on gfx942 and 128 on gfx950). Sparge mask at BLOCK_M=256, BLOCK_N=KV tile with `pad_block_divisible=True`, BHSD to BSHD permute, then sparse MHA v4. On gfx942 anything other than native FP8/FP8 or INT8/FP8 is rejected. MXFP8 uses `mha_v4_mxfp8` with `block_mask` when that helper accepts it; otherwise packed plus `block_attn_mask_to_ragged_lut`.
- `xfuser/core/distributed/runtime_state.py`: treat Sparge-MHA-v4 like other low-precision backends; fail at init if `mha_v4` has no `block_mask`, if the GPU is neither gfx950 nor gfx942, or if a gfx950-only recipe is selected on gfx942.
- `xfuser/model_executor/layers/usp.py`: include the new backends in Ulysses head-balance (same cost sink as Triton/Flex Sparge).
- `xfuser/model_executor/models/runner_models/base_model.py`: fold them into `_SPARGE_ATTENTION_BACKENDS` so existing cross-attn validation applies.
- `README.md`: backend table (the eight new names, plus rows for the pre-existing `aiter_sparge`, `aiter_sparge_v2`, and `flex_block_sparge` that the table was missing), arch/tile paragraph, and hybrid example.
- `tests/test_aiter_sparge_mha_v4.py`: registry, mask plumbing for both `mha_v4` and `mha_v4_mxfp8`, per-arch tile geometry, gfx942 recipe reject, causal/dropout rejects, cosine vs dense sibling.
- `tests/test_aiter_mixed_attention.py`: the dense `AITER_I8FP8` case now also runs on gfx942, and the kernel probe looks under the `MI300/` subdirectory there.

CLI names (enum `AITER_*_SPARGE`):

`aiter_i8fp8_sparge`, `aiter_fp8_sparge`, `aiter_mxfp8_sparge`, `aiter_f8f6_sparge`, `aiter_mxfp6_sparge`, `aiter_f6f4_sparge`, `aiter_mxfp4_sparge`, `aiter_f4f4_sparge`

Call path: `_build_sparge_block_mask` -> permute -> `mha_v4` / `mha_v4_mxfp8` with `block_mask`. The work table is not a Python argument; AITER derives it from the mask. The one exception is the MXFP8 fallback for AITER builds whose `mha_v4_mxfp8` predates `block_mask`, which builds the ragged LUT here and passes it to `mha_v4_packed`.

MHA v4 sparse kernels and AITER already support GQA (power-of-two Q/KV head ratio up to 16; LUT rows are one per query head). These Sparge backends still require matching Q and KV head counts because `_build_sparge_block_mask` scores Q and K with the same head index. Models that expand KV before USP (for example Krea-2) already arrive as MHA. True GQA into this Sparge path is not wired here.

## Usage

Hybrid (`--attention_backend` must be left unset when `--use_hybrid_attn_schedule` is on):

    --use_hybrid_attn_schedule
    --hybrid_attn_high_precision_backend aiter_fp8
    --hybrid_attn_low_precision_backend aiter_fp8_sparge
    --num_hybrid_attn_high_precision_steps N
    --cross_attention_backend aiter

Same pattern for the other seven recipes (`aiter_mxfp4` / `aiter_mxfp4_sparge`, and so on) on gfx950.

Do not set `--cross_attention_backend` to any Sparge backend (`aiter_fp8_sparge`, `aiter_sparge`, `flex_block_sparge`, etc.). Cross-attn is already rejected when Sparge is in play.

## Guards / non-goals

Explicit `NotImplementedError` / `RuntimeError` (no silent SDPA fallback):

- gfx950 or gfx942; on gfx942 only the native FP8/FP8 and INT8/FP8 recipes
- head dim 128
- matching Q/KV head counts at this layer (Sparge mask builder, not the kernel)
- non-causal, dropout_p == 0
- no varlen `indices_k`
- no LSE, so no ring attention on these backends

Left alone: `aiter_sparge`, `aiter_sparge_v2`, `flex_block_sparge`, SSTA sparse sage.

## Tests

    pytest tests/test_aiter_sparge_mha_v4.py tests/test_aiter_mixed_attention.py

On gfx950 with a sparse-capable AITER this is 51 passed, including all eight Sparge-vs-dense siblings (all-true Sparge thresholds, cosine > 0.95) plus the mixed-attention dense suite.

These tests need a host where AITER imports successfully; the format constants the backends pass to `mha_v4` only exist then. GPU tests skip if `mha_v4` lacks `block_mask`, if the arch does not support the recipe, or if the matching `fwd_hd128_*_sparse.co` is missing.

## Dependencies

- AITER `mha_v4` / `mha_v4_packed` with `block_mask` (and `mha_v4_mxfp8` with `block_mask` for the MXFP8 fast path)
- `aiter.ops.mha_v4.mha_v4_kv_tile()` when present; otherwise the KV tile is inferred from the arch
- Sparse code objects under `hsa/gfx950/fmha_v4_fwd/` and, for gfx942, `hsa/gfx942/fmha_v4_fwd/MI300/`


xfuser/core/distributed/attention_backend.py
```diff
@@ -24,6 +24,12 @@
 logger = init_logger(__name__)
 
 ATTENTION_FUNCTION_REGISTRY = {}
+_AITER_MHA_V4_AVAILABLE = False
```
jcaraban: a bit too fine-grained? can it be compacted into fewer availability checks?

xfuser/core/distributed/attention_backend.py
```diff
@@ -588,6 +620,14 @@ class AttentionBackendType(Enum):
...
+    AITER_I8FP8_SPARGE = "AITER I8FP8 Sparge"
+    AITER_FP8_SPARGE = "AITER FP8 Sparge"
+    AITER_MXFP8_SPARGE = "AITER MXFP8 Sparge"
+    AITER_F8F6_SPARGE = "AITER F8F6 Sparge"
+    AITER_MXFP6_SPARGE = "AITER MXFP6 Sparge"
+    AITER_F6F4_SPARGE = "AITER F6F4 Sparge"
+    AITER_MXFP4_SPARGE = "AITER MXFP4 Sparge"
+    AITER_F4F4_SPARGE = "AITER F4F4 Sparge"
     AITER_SAGE = "AITER Sage"
     AITER_SPARSE_SAGE = "AITER Sparse Sage"
```
jcaraban, line 641: shall this PR prune the outdated Triton sparges?

README.md
```diff
@@ -284,10 +284,21 @@ Several different attention backends are supported:
 | [AITER F6F4](https://github.com/rocm/aiter) | aiter_f6f4 |
 | [AITER MXFP4](https://github.com/rocm/aiter) | aiter_mxfp4 |
 | [AITER F4F4](https://github.com/rocm/aiter) | aiter_f4f4 |
+| [AITER I8FP8 Sparge](https://github.com/rocm/aiter) | aiter_i8fp8_sparge |
```
jcaraban: maybe compact the dense / sparge rows into a single dense / sparse line listing the variants?

xfuser/core/distributed/attention_backend.py
nsakkine: Slightly too fine-grained, but felt that mostly needed. Refactored into a capabilities dataclass. Hopefully better?

xfuser/core/distributed/attention_backend.py
nsakkine: Pruning these would break some workloads, like Hunyuanvideo 1.5 distilled sparse, which required different tile sizes. Proposing to not do it

README.md
nsakkine: Better?

jcaraban (approved):

avjves (approved):
LGTM! This AITER version probing is messy though. I get that it's pretty much needed if we want to support bleeding edge stuff here, but we should aim to remove those ASAP and just add a minimum AITER version instead. I think AITER gets bumped every 2 weeks or so.

PR #750, opened 2026-08-13
DistVAE updates including tile-parallel VAE execution

pds-amd:
## Summary

When VAE tiling and `--use_parallel_vae` are active on a supported VAE, DistVAE assigns complete tiles to ranks and assembles the decoded output afterward. Without tile distribution, the decoder is sharded along its latent-height axis.

VAE tiling primarily reduces peak memory. Complete-tile distribution can also reduce decode latency when the tile grid supplies enough balanced work.

Tiling also has a nice side effect of shorter compile times.

Some illustrative benchmark examples are provided below.

## Execution modes

| Configuration | Decoder execution |
| --- | --- |
| `--use_parallel_vae` | Row-sharded decode |
| `--enable_tiling` | Local tiled decode |
| Both flags, supported VAE | Complete tiles distributed across VAE ranks |
| Exact tile size or overlap flags | Enables tiling and uses the requested output-pixel geometry |

Models declaring `use_parallel_vae_encoder=True` also receive a row-sharded encoder. Complete-tile distribution applies only to decoding.

Row sharding splits one latent along its height axis. It can reject shapes that do not align with the VAE's patch or downsampling unit, leave too few bands for the selected rank count, or cannot provide the required convolution halo. Complete-tile distribution avoids those intra-tile split constraints and can therefore use configurations that row sharding rejects.

## Included

- Add `VAEManager` to configure every unique VAE used by single-stage and multi-stage pipelines.
- Add exact output-pixel tile controls:
  - `--vae_tile_size_height`
  - `--vae_tile_size_width`
  - `--vae_tile_overlap_height`
  - `--vae_tile_overlap_width`
- Preserve rectangular tile geometry and validate it against the selected execution mode.
- Support in-process `GroupCoordinator` groups and dedicated raw `ProcessGroup` instances.
- Apply channels-last conversion and VAE options once per VAE and in lifecycle order.
- Restore Torch GroupNorm before DistVAE adapter discovery when AITER replaced it.
- Report actionable shape, padding, and OOM diagnostics. Decoder row-split failures suggest complete-tile decoding when supported.
- Require the DistVAE beta9 public API instead of importing adapter implementations directly.

## Dependency

Depends on the companion DistVAE beta9 change. https://github.com/xdit-project/DistVAE/pull/14

DistVAE beta9 must be published before this PR merges. The latest version currently available from PyPI is beta5, while `setup.py` requires `distvae>=0.0.0beta9`.

### gfx1201

[...]

PR #759, opened 2026-08-24
LTX-2.5 optimizations: add parallel VAE decoding and AITER RMSNorm

jjuvonen-amd:
# Summary

LTX-2.5 diffusion VAE decoder is parallelized on N-GPU runs when `--use_parallel_vae` is enabled, and `diffuser.models.normalization.RMSNorm` layers are swapped to faster AITER's CK-Tile `rms_norm` when available.

# Changes

**1. Tile-parallel diffusion VAE decode (`--use_parallel_vae`)**
New file: `xfuser/model_executor/layers/ltx2/diffusion_decoder.py`

`xFuserLTX2VideoDiffusionDecoderWrapper` subclasses `LTX2VideoDiffusionDecoderModel` and overrides `tiled_decode`. The decoder's existing `tiled_decode` already decomposes the work into an independent (temporal × height × width) tile grid where each tile runs `forward_stage_4` + denoise (8 neighborhood-attention blocks) with no cross-tile dependency during compute. The wrapper distributes those tiles across the SP group via round-robin ownership:

- All ranks run `forward_stages_1_to_3` on the full volume (cheap; avoids communicating the intermediate feature tensor)
- Each rank computes only its assigned tiles
- One `all_reduce(SUM)` of a `[n_tiles, 3]` shape-metadata tensor, then per-tile `broadcast(src=owner)` to deliver every tile to all ranks
- All ranks run the stock blend/assembly loop on the gathered tiles

For the default 1024×1536×121 config this produces 12 tiles, adequate for 2/4/8 GPUs. Noise determinism: the shipping 1-step x0 path uses per-tile seeding (`base_seed XOR f(t,h,w)`), which is rank-invariant without requiring non-owners to know tile shapes in advance. Multi-step paths draw a shared noise canvas identically on all ranks.

Activated by `--use_parallel_vae`. Automatically implies `enable_tiling()` on the decoder (required for `tiled_decode` dispatch). Falls back to the stock single-GPU `tiled_decode` when `sp_world_size == 1` or the tile count is less than 2.

Note: [PR #750's](https://github.com/xdit-project/xDiT/pull/750) DistVAE tile-parallel mechanism does not cover LTX-2.5. The diffusion decoder uses pure `nn.Linear` + pixel-shuffle + 3D neighborhood-attention blocks, for which DistVAE has no adapter. This wrapper is the only viable multi-GPU path for this decoder.

**2. AITER block RMSNorm acceleration**

New file: `xfuser/model_executor/layers/norms.py`

`diffusers.models.normalization.RMSNorm` is a Python implementation that up-casts to float32 before normalizing; AITER's CK-Tile `rms_norm` kernel operates natively in bfloat16. `norms.py` provides:

- `_aiter_rms_norm`: `torch.library.custom_op` wrapper fixing a false mutation declaration in AITER's op schema (mutates_args=()) that would otherwise cause `torch.compile` to skip CUDA graph capture at every call site
- `_AITERRMSNorm`: `nn.Module` drop-in for `diffusers.models.normalization.RMSNorm`; handles both `elementwise_affine=True` and `=False` (ones buffer for the no-weight case)
- `_replace_rms_norms_with_aiter(model)`: walks any model and replaces every `diffusers.models.normalization.RMSNorm` instance with `_AITERRMSNorm`, leaving `torch.nn.RMSNorm` (the C++ implementation, already faster than AITER) untouched
- Called in `xFuserLTX2VideoTransformer3DWrapper.__init__` before `model.to(device)` so replacement buffers are placed correctly by the device move. LTX-2.5 has 384 such block norms (8 per block × 48 blocks), each an instance of the slow Python variant. (saves ~13ms/step)

The utility is model-agnostic. Any future model using `diffusers.models.normalization.RMSNorm` for block norms can call `_replace_rms_norms_with_aiter` directly.

**3. File reorganization**

LTX-2.5-specific layer files moved under `xfuser/model_executor/layers/ltx2/`:

- `layers/ltx2_na3d_eager_attn.py` renamed to `layers/ltx2/na3d_eager_attn.py`
- New: `layers/ltx2/diffusion_decoder.py`

# Tests

LTX-2.5 Output quality remains similar after parallel-vae or AITER RMSNorm.

## Original 1-GPU (MI350)

https://github.com/user-attachments/assets/8338b994-20a2-42b3-909b-18c5dbac59aa

## 8-GPU (MI350) parallel-VAE, AITER RMS-Norm

https://github.com/user-attachments/assets/5a96961d-18a2-437d-afc5-cf85c742562d


xfuser/model_executor/models/runner_models/ltx.py
kTorp: nitpick but is it supposed to say "Triton" here? Other comments mentions tiled PyTorch SDPA as the fallback

jjuvonen-amd: Good catch, I fixed the log message to use "tiled PyTorch SDPA".

kTorp (approved):

avjves (approved):

PR #764, opened 2026-08-26
Graph capture profiling support

kyle-hoffmeyer:
## Summary

Adds graph capture profiling support for xDiT workloads. This is needed to capture per-kernel shapes and call stack information from workloads which is only available during the graph capture phase. This information is needed to perform roofline analysis on graph-replay workloads.

**`xfuser/config/args.py`**: Adds `profile_capture_phase: bool = False` field to `xFuserArgs` (xDiT's main config class) and `--profile_capture_phase` CLI flag.

**`xfuser/model_executor/models/runner_models/base_model.py`**: Added `_run_compile_warmup()` method to handle graph capture profiling logic.

Call stack before: `_compile_model() -> _run_timed_pipe()` 
Call stack with patch: `_compile_model() -> _run_compile_warmup()  -> _run_timed_pipe()`  

When `--profile_capture_phase` is enabled and conditions are met, this method splits the compile warmup into two phases:

```python
def _run_compile_warmup(self, compile_args: dict) -> None:
    enable_capture = (
        self.config.profile_capture_phase
        and get_world_group().rank == 0
        and self._get_compile_mode() == "reduce-overhead"
        and not self.pipe.scheduler.config.shift_terminal
    )
    if not enable_capture:
        self._run_timed_pipe(compile_args)
        return

    # Phase 1: compile + eager warmup (1 step, no profiler)
    compile_args["num_inference_steps"] = 1
    self._run_timed_pipe(compile_args)

    # Phase 2: CUDA graph recording (1 step, with profiler)
    with profile(record_shapes=True, with_stack=True) as capture_prof:
        self._run_timed_pipe(compile_args)
    capture_prof.export_chrome_trace(...)
```

#### `enable_capture` conditions
- **`self.config.profile_capture_phase`**: The `--profile_capture_phase` CLI flag must be explicitly set.
- **`get_world_group().rank == 0`**: Only rank 0 collects the capture trace, matching vLLM/SGLang behavior.
- **`self._get_compile_mode() == "reduce-overhead"`**: Capture profiling is only needed for models compiled with CUDA graphs (`reduce-overhead` mode). In `default` mode, there are no CUDA graphs — the profiler already sees individual kernel launches during the replay trace, so no separate capture phase is needed.

- **`not self.pipe.scheduler.config.shift_terminal`**: Models whose diffusers scheduler sets `shift_terminal` cannot run with `num_inference_steps=1` which is required for graph capture profiling. Bug tracked https://github.com/huggingface/diffusers/issues/14411.
#### Why the warmup is split into two phases

In xDiT, `torch.compile(mode="reduce-overhead")` uses a three-phase warmup:
1. **Phase 1** (eager warmup): Dynamo traces the forward pass, Inductor compiles and generates kernels, runs eagerly.
2. **Phase 2** (CUDA graph recording): The compiled kernels are recorded into a CUDA graph, capturing individual kernel launches.
3. **Phase 3+** (CUDA graph replay): A single `cudaGraphLaunch` replays all kernels — individual launches are invisible to the profiler.

**`xfuser/model_executor/models/runner_models/ltx.py`**: Both `xFuserLTX23VideoModel` and `xFuserLTX2VideoModel` have their own `_compile_model()` overrides. These are updated to call `_run_compile_warmup()` instead of `_run_timed_pipe()`.


kyle-hoffmeyer:
@avjves thank you for your feedback:

"The rank 0 check makes multi-GPU runs desync. Rank 0 will run for just 1 inf step whereas the rest will run for 2 or more, depending on the model."

Rank 0 runs `run_timed_pipe` with num_inference_steps=1 twice. The other ranks run `run_timed_pipe` with num_inference_steps=2 once. From my understanding, for the denoising steps, this shouldn't be an issue since it's running two denoising steps either way. This would mostly be an issue for collective operations that are ran once for every call to `run_timed_pipe` rather than for every denoising step, correct? Since rank 0 would be running two collectives but the rest of the ranks will be running just one? If you could confirm my understanding on this and provide more details that would be useful.

"We run the compile for 2 or more inf steps so the compile sees all operations. How does this 1 inf step graph recording work with Wan2.2 for example that changes it's internal transformer ~half way into the process."

Wan models run `torch.compile()` in default mode so it doesn't run in graph mode. That being said, this is an important limitation of the graph capture profiling. Currently, all of the xDiT models that run in graph mode only have one graph and the profiling assumes that this will stay true. But if Wan2.2 were to change in the future to run in graph mode, this would raise issues, as it would create two graphs for the two different transformers.
We've explored methods of explicitly profiling the graph capture phase, but this is difficult as it happens through torch Inductor which isn't patchable from within xDiT. If you have suggestions for how to more explicitly profile the graph capture phase that would be very useful.

"All models should probably use the new wrapper?"

Most models derive `_compile_model` which calls `run_timed_pipe`/`run_compile_warmup` from `base_model.py`. LTX defines it's own _compile_model and runs in graph mode which is why I'm modifying it to call the new wrapper function. The other models either derive `_compile_model` from `base_model.py` or they don't currently run in graph mode. If it would be useful to make all the other models call the wrapper in case graph mode becomes supported for these models, I can make this change.

avjves:
Thanks for this :)

Replying to replies 😄:
Running inf_steps=1 twice in row is not equivalent to running with inf_steps=2. The compiler still sees new operations during the second step (e.g, the change between two inf steps), but the bigger problem is that it runs text encoder, VAE etc. twice. We often want to parallelize the VAE as well, so this would just cause an indefinite hang up there, right?

I'll be 100% honest, I'm not too familiar with this graph capture phase, so I might be missing something.  Also, for some models (such as Wan2.2), we don't run 2 inference_steps during compile but the full amount of steps (generally 40).

So that nothing breaks, I think we should always run the same code on all ranks in play. Currently we output traces for each rank, e.g "profile_rank_0.json.gz", "profile_rank_1.json.gz" and so forth. Would this not work for the graph capture phase as well? Does it _have_ to only profile the rank0?


For the other ones:
Ah, okay, good. Yes I forgot this is limited to `reduce-overhead`. It would probably be better to explicitly disable the support for said X models, as right now if I were to run a model that runs with "default" mode, it'll just silently _not_ produce the capture phase trace.

You can disable/enable it via the ModelCapabilities list. If you add "profile_capture_phase": False in the `base_model.py`, then you can per-model add "profile_capture_phase": True within each model file to enable the support for the models that support this. If one tries to use the flag with unsupported model, it will crash explaining that.

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -745,16 +746,24 @@ def profile(self, input_args: dict) -> Tuple[DiffusionOutput, list, torch.profil
             active=self.config.profile_active,
         )
         num_repetitions = self.config.profile_wait + self.config.profile_warmup + self.config.profile_active
+        with_stack = self.config.profile_capture_phase
+        batch_size = input_args.get("num_images_per_prompt", 1)
```
avjves: We have `--batch-size` flag as well. Should this use that instead?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -745,16 +746,24 @@ def profile(self, input_args: dict) -> Tuple[DiffusionOutput, list, torch.profil
...
+        height = input_args.get("height", 0)
+        width = input_args.get("width", 0)
+        if self.config.profile_capture_phase:
```
avjves: I think it's actually fine to always annotate it with `execute_diffusion_{batch_size}_{height}x{width}` :) Just remove the `model_inference` annotation

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -745,16 +746,24 @@ def profile(self, input_args: dict) -> Tuple[DiffusionOutput, list, torch.profil
             active=self.config.profile_active,
         )
         num_repetitions = self.config.profile_wait + self.config.profile_warmup + self.config.profile_active
+        with_stack = self.config.profile_capture_phase
```
avjves: Could we add a new CLI parameter `--profile-with-stack`? This would be useful for other usages as well. I think this param should require `--profile` flag or at least leave a log saying that it's not enabled -> no profiles outputted.

kyle-hoffmeyer:
@avjves, thank you for your feedback.
- I've added a `--profile_with_stack` flag. I've verified that a warning is logged if a user sets this to true without setting `--profile`.
- `batch_size` is set to `self.config.batch_size` if it's set; otherwise 1
- Replaced default annotation with new one
- All ranks are profiled during graph capture phase and graph replay phase instead of just rank 0. Verified that every rank contains a corresponding capture and replay trace.
- Updated ModelCapabilities to have profile_capture_phase False by default. Overwrote this in Flux and LTX ModelCapabilities to support profile_capture_phase being True. Verified that profiling capture phase works for these models. Verified that the workload fails for other models when profile_capture_phase is set.
- Verified that workload runs E2E and profiling works when `--use_parallel_vae` is set.

Please let me know what you think. Thanks!

avjves (approved):
Looks good to me now :) Thanks, merging!

PR #765, opened 2026-08-27
fix(flydsl): select QK RoPE block size from device wave size

pds-amd:
Use the device-reported warp size when selecting the fused QK RMSNorm/RoPE kernel configuration. This prevents wave64 shuffle lowering on wave32 RDNA4 GPUs, which produced corrupted FLUX and Qwen images.

Fall back to the reference implementation when the reported wave size is missing, invalid, or incompatible with the head dimension.

Without this on gfx1201 we just get noise images because 64 is not valid.


avjves (approved):
Thanks! 🙏

PR #766, opened 2026-08-27
fix(flydsl): use device wave size for WAN QK RoPE fusion

pds-amd:
Extract device wave-size detection into a shared FlyDSL utility and use it across the FLUX, Qwen, and WAN fused QK normalization/RoPE kernels.

Select WAN block geometry from the hardware-reported wave size, preventing wave64 shuffle reductions on wave32 RDNA4 devices. Fall back to the reference implementation when the wave size or tensor geometry is unsupported.

### DIFF
--- a/xfuser/model_executor/layers/flydsl_utils.py
+++ b/xfuser/model_executor/layers/flydsl_utils.py
@@ -1,0 +1,11 @@
+from typing import Optional
+
+import torch
+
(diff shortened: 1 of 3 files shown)


avjves (approved):
LGTM! Thanks! I'll update the Z-Image PR to use this :)

PR #770, opened 2026-09-01
Migrate to `pyproject.toml`

nsakkine:
## Summary

Migrates packaging from `setup.py` to `pyproject.toml` and uses setuptools-scm to derive versions from git tags and write them to `xfuser/__version__.py` at build/install time.

- Add `pyproject.toml` with project metadata, dependencies, optional extras, and `xdit` entry point
- Remove `setup.py`
- Configure setuptools-scm (`fallback_version`, `write_to_source`, simple version template)
- Update publish workflow: `fetch-depth: 0` for tag-based versions, `setup-python@v5`
- Ignore `xfuser/__version__.py` (generated file, untracked from git)
- Update references in tests, README, Dockerfile, and compat/docs comments

## Test plan

- [x] `python -m build` produces wheel/sdist with correct version from git tags
- [x] `pip install -e .` generates `xfuser/__version__.py`
- [x] `xdit` console script entry point is present in the wheel


pyproject.toml
```diff
@@ -0,0 +1,73 @@
...
+name = "xfuser"
+description = "A Scalable Inference Engine for Diffusion Transformers (DiTs) on Multiple Computing Devices"
+readme = "README.md"
+requires-python = ">=3.10"
+authors = [
+    { name = "xDiT Team", email = "fangjiarui123@gmail.com" },
+]
+dynamic = ["version"]
+dependencies = [
+    "torch>=2.4.1",
```
avjves, line 15: These deps are pretty old - we should probably update these at some point. Torch 2.4, I have a feeling something _might_ break if we used that one 😅 

But maybe not necessary in this PR - let's just get the versioning running.

avjves (approved):
LGTM! Left a note for a future PR / house keeping, but will merge as-is.

PR #771, opened 2026-09-01
Fix `DiffusionOutput` to not wrap `videos=None` into a list

Arech8:
`class DiffusionOutput()` constructor sets `None` as default value for `videos` parameter. Few lines later it wraps it into a list turning `videos=None` into `videos=[None]`. Almost all other code of the project tests whether videos field is set with `if obj.videos:` and `bool( [None] ) == True`, so the code takes the `if` branch which isn't meant for that. Leaving `videos` set to `None` prevents that.


avjves (approved):
LGTM!

PR #778, opened 2026-09-09
Add determinism check

Arech8:
This PR adds two CLI/config arguments: integer `--determinism_check` and string `--determinism_check_report_ranks`. When `--determinism_check` value is positive, this enables deterministic inference check in the xFuser: in the inference loop, on the first iteration, the output a model produces is saved (on each rank independently), and is compared to what the model returns on each next iteration. Whenever the outputs mismatch, an error message is produced, and if the total number of checks failed to the moment is less than or equal to `int(determinism_check)`, the baseline and the differing outputs are dumped to the same output directory for further inspection. Hence a value of `--determinism_check` enables the check and lets the user to limit the number of dump files (since usually it's enough to have a single evidence of non-deterministic behavior).

Value of `--determinism_check_report_ranks` helps with limiting the number of output dumps by letting a user to restrict outputs of which ranks are dumped to files.

Additionally, an independent `determinism_check_results()` helper method is provided to enable detection of determinism check failures for external model runners. The helper should stay in the repo to maintain coupling between certain format of error messages and dump file names, which is enforced by tests.

### DIFF
--- a/xfuser/config/args.py
+++ b/xfuser/config/args.py
@@ -1,6 +1,7 @@
-import sys
 import argparse
 import dataclasses
+import os
+import sys
 import warnings
 from dataclasses import dataclass
 from typing import Optional, List, Tuple, Union
@@ -67,6 +68,72 @@ def nullable_str(val: str):
     if not val or val == "None":
         return None
     return val
+
+
+_DETERMINISM_CHECK_HELP = (
+    "Set to a positive failure threshold to compare every timed iteration with "
+    "the first using exact equality, not a tolerance. The failure handler is "
+    "called for each rank enabled by --determinism_check_report_ranks on each "
+    "divergence until the threshold is reached. "
+    "Disabled at 0 or below. Enabling retains and compares full output payloads, "
+    "which adds memory, synchronization, and transfer overhead."
+)
+
+_DETERMINISM_CHECK_REPORT_RANKS_DEFAULT = "last"
+
+_DETERMINISM_CHECK_REPORT_RANKS_HELP = (
+    "Ranks on which determinism failures save serialized outputs. Accepts a "
+    "comma-separated list of integers, an empty value, or one of: all, first, "
+    "last, none. Empty and none select no ranks. "
+    f"Defaults to {_DETERMINISM_CHECK_REPORT_RANKS_DEFAULT}. Many failures "
+    "may produce many large output files; restrict output to selected ranks "
+    "based on model specifics. The none value is useful with a positive "
+    "--determinism_check when only log messages, but no files, are needed."
+)
+
+
+def _normalize_determinism_check_report_ranks(value: str) -> frozenset[int]:
+    """Resolve a rank string; ``none`` and empty values select no ranks."""
+    try:
+        world_size = int(os.environ.get("WORLD_SIZE", "1"))
+    except ValueError as exc:
+        raise ValueError("WORLD_SIZE must be a positive integer") from exc
+    if world_size <= 0:
+        raise ValueError("WORLD_SIZE must be a positive integer")
+    if not isinstance(value, str):
+        raise TypeError("determinism_check_report_ranks must be a string")
+
+    value = value.strip()
+    if value == "all":
+        ranks = frozenset(range(world_size))
+    elif value == "first":
+        ranks = frozenset({0})
+    elif value == "last":
+        ranks = frozenset({world_size - 1})
+    elif value in ("none", ""):
+        ranks = frozenset()
+    else:
+        parts = [p.strip() for p in value.split(",")]
+        if any(not part for part in parts):
+            raise ValueError(
+                "determinism_check_report_ranks must be a comma-separated list "
+                "of integers, empty, or one of: all, first, last, none"
+            )
+        try:
+            ranks = frozenset(int(part) for part in parts)
+        except ValueError as exc:
+            raise ValueError(
+                "determinism_check_report_ranks must be a comma-separated list "
+                "of integers, empty, or one of: all, first, last, none"
+            ) from exc
+
+    invalid_ranks = sorted(rank for rank in ranks if rank < 0 or rank >= world_size)
+    if invalid_ranks:
+        raise ValueError(
+            "determinism_check_report_ranks contains ranks outside "
+            f"0..{world_size - 1}: {invalid_ranks}"
+        )
+    return ranks
 
 
 @dataclass
@@ -154,6 +221,8 @@ class xFuserArgs:
     fp8_precision_override_suffix_patterns: Optional[str] = None
     # Model runner specific
     num_iterations: int = 1
+    determinism_check: int = 0
+    determinism_check_report_ranks: str | frozenset[int] = "all"
     profile: bool = False
     profile_capture_phase: bool = False
     profile_with_stack: bool = False
@@ -205,6 +274,11 @@ class xFuserArgs:
     distilled_transformer_2_path: Optional[str] = None
 
     def __post_init__(self):
+        self.determinism_check_report_ranks = (
+            _normalize_determinism_check_report_ranks(
+                self.determinism_check_report_ranks
+            )
+        )
         if self.profile_with_stack and not self.profile:
             logger.warning(
                 "--profile_with_stack has no effect without --profile; "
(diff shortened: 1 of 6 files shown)


xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -698,6 +762,26 @@ def run(self, input_args: dict) -> Tuple[DiffusionOutput, list]:
                 timings.append(timing)
                 log(f"Iteration {iteration + 1} completed in {timing:.2f}s")
 
+            if self.config.determinism_check > 0:
```
avjves, line 776: This now affects the timings as it's ran between the `inference_start` and `inference_end` calls. We should probably emit some logs to tell user to _not_ trust the total timing even if no determinism check fails, as it it's now inflated.

avjves: Is the desired use-case here to keep this enabled at all times to spot possible deviations or to run it as one-off tests at some intervals?

I'm just thinking if it would make sense to have it's own path and not hijack `run()`. In the latter case it would possibly make sense.

xfuser/config/args.py
```diff
@@ -69,6 +70,68 @@ def nullable_str(val: str):
     return val
 
 
+_DETERMINISM_CHECK_HELP = (
+    "Set to a positive failure threshold to compare every timed iteration with "
+    "the first using exact equality, not a tolerance. The failure handler is "
+    "called on the threshold-reaching divergence and every later divergence. "
```
avjves: Isn't this now reversed?
Failure handler is called for every divergence _until_ the threshold is reached 🤔  Or am I reading it wrong?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -668,13 +672,73 @@ def _compile_model(self, input_args: dict) -> None:
             compile_args["num_inference_steps"] = warmup_steps
         self._run_compile_warmup(compile_args)
 
+    def _save_determinism_check_failed_outputs(
```
avjves, line 675: This now re-implements the saving function that some models can override. MiniMax, for example, does so. Could this just re-use that function?

xfuser/core/utils/determinism_check_results.py
```diff
@@ -0,0 +1,170 @@
+"""This module is meant to be an independent collection of utilities for handling determinism check
+results, primarily for use by external higher-level runners. It's meant to couple with the xFuser
+only by externally visible artifacts it produces, such as logs or files. However, dur to this
```
avjves: typo: dur -> due

xfuser/model_executor/models/runner_models/base_model.py
Arech8: Thank you for the review, Aleksi!

That's right, it affects `inference_start`/`inference_end` indeed, but their only use is to output "Total time spent". The measurement doesn't go anywhere else, so I'm not sure how serious is the issue. Does anyone use that measurement manually?

The idea with the log is really proper one, I'll add it in the next push, thanks!

Ideally, this check should always be on, since it's crucial to detect non-deterministic behavior, and the math is simple: the more runs covered - the more chances to detect it. But I'm not exactly sure if everyone would be happy to always have it on...

Why are you concerned about "hijacking `run()`"? If the check is disabled, the cost of everything is a condition evaluation and a jump, which is in Python world typically considered as "literally nothing". What makes you concerned about that?

xfuser/config/args.py
Arech8: Oh, right, there was a bug in the activation condition and my agent did a silly thing... I should have noticed that, I'm sorry. Of course, the description is wrong. The handle is called until the threshold is reached.
Will fix this, thanks for spotting!

xfuser/model_executor/models/runner_models/base_model.py
Arech8: You're talking about `save_output()`, right?
Yeah, I forgot about overrides...

Hmm.... The reason this function exist is (a) dump files need different naming, and (b) to save raw data with `torch.save()`.

How can we achieve that with `save_output()`?

For (a), as long as derived models don't override `self.get_output_name()`, we can hijack it to generate proper file name from (also hijacked) `pipe_args` of the `DiffusionOutput` instance - we could add special fields there that amend `get_output_name()` behavior, and then cleanup the fields, so this won't affect anything else. Seems a bit hacky, but this would work.

(b) is a bigger issue... So far I had no use in raw dumped tensors, but I'd imagine they are going to be useful for pinpointing changes that aren't visible visually from compressed outputs. Though, not sure how exactly useful, since check failure is enough to prove divergence and what one would do with the rest?.... Dunno, perhaps we can skip it and when someone need that, they could live-patch the code... Would be nice to have a good solution though...

Wdyt?

xfuser/model_executor/models/runner_models/base_model.py
Arech8: log message added

xfuser/config/args.py
Arech8: fixed this, thanks!

xfuser/core/utils/determinism_check_results.py
Arech8: fixed, thanks!

xfuser/model_executor/models/runner_models/base_model.py
Arech8: All right, I guess we could ditch (b) for now: a single Wan2.2 clip we have weights around 1Gb as `torch.save()`, so it's infeasible to keep them all and we probably just shouldn't save them by default at least. (b) could be added when really needed on top of `save_output()` call, but for now let's simply forget about it.

So, yeah, I could reuse `save_output()`. What do you think of the proposed method - modify `pipe_args` to change the behaviour of ``get_output_name()`? At least currently, `get_output_name()` isn't overridden and I could add a hint there that the additional behaviour must be preserved in the override.

xfuser/model_executor/models/runner_models/base_model.py
avjves: That is true. We at least do not use it, but someone else might. I'd say fairly low chance, though.

Logic-wise there's no issue of course - without the flag it's dead code. I'm mostly thinking about this from code angle, where this now breaks SRP even further and makes the code a bit harder to understand (i.e "Hm, do I need to understand what's going on here" when a new guy looks at the main `run()` function). Maybe it can be cleaned up a bit by not having the logic within it, just a call to somewhere else?

xfuser/model_executor/models/runner_models/base_model.py
avjves: Nice, sounds good!

I don't see a reason why a model _would_ override the function. The possiblity is there, though 😅 

Hmm, yeah, it's not going to be the nicest thing without some bigger refactoring. One option, also quite hacky, is for you to hijack the `get_output_name()` directly in the `save_determinism_check_failed_outputs()`, add your desired changes on top of the current output name, call save and revert to be on the safe side. This would require no new fields etc and save function would work just fine, I think.

Arech8: Hmm.. Right, that might be actually a cleaner more scoped solution. It's way slower in runtime, but I think we can deal with it, given it should typically happen no more than once or twice during the whole inference.

xfuser/model_executor/models/runner_models/base_model.py
Arech8: ah, right, that's good idea indeed, I'll change it. Thanks!

Arech8: refactored that in a separate function.
The condition still inside the loop, - otherwise the runtime cost will also have function call added, and...better avoid that.

xfuser/model_executor/models/runner_models/base_model.py
Arech8: Implemented with `save_output()` and hijacking of `get_output_name()`.

Arech8: ah, also made `last` the default for `..._ranks` filter, so it won't make too much of an output.

avjves (approved):
Thanks, it looks great now 🙏

PR #779, opened 2026-09-09
new MHA v4 features, bugs, fixes

jcaraban:
- Add two backends — `aiter_bf16` (BF16 Q/K/V) and `aiter_bf16fp8` (BF16 Q/K + FP8 V), routed through mha_v4; `aiter` keeps using the original MHA v3.
- Move MXFP8 onto the generic scale-mode API — AITER deprecated mha_v4_mxfp8 and warns from it, and Dynamo can't trace warnings.warn, so every MXFP8 path broke under fullgraph=True. The capability probe now detects the generic API once; the legacy entrypoint is used only on older builds.
- Reject varlen packed keys on dense MHA v4 — all eight dense backends accepted attention_kwargs and silently dropped it, folding padded keys into the softmax denominator. Measured on an LTX-2 cross-attention shape: cosine 0.99962, max abs error 0.29. LTX-2 had no runner guard and would have taken that path.
- Keep the v3 FP8 dense path opaque to Dynamo — _aiter_fp8_dense_attention was a plain function while its varlen sibling was already a custom op, so compilation traced into AITER and hit a `Tensor | None` annotation. Same custom-op boundary now; fixes Ideogram-4 at head dim 256 with no AITER change.
- Skip/xfail recipes current AITER can't serve — aiter_mxfp4 has no dense kernel temporarily (skipped, probed per build); dense MXFP4 V is numerically wrong at ragged sequence lengths (non-strict xfail, flips to XPASS when fixed).
- Result — attention suite goes 91 passed/12 failed → 146 passed / 0 failed / 8 skipped / 2 xfailed. core identical to an origin/main control worktree, so no regressions.


avjves (approved):
LGTM, but we really need to do something about these attention backends soon 😅 It's just expanding and expanding.

PR #780, opened 2026-09-11
FastH3 VSA-H3 T2VA support

hakankiymaz-amd:
## Summary

FastH3 was trained on 64-token mixed-modality tiles with prefix-exempt 90% video sparsity. Wan/Jenga 128-block VSA does not implement that geometry, so this PR adds a MiniMax-H3-only VSA-H3 path instead of extending the existing VSA kernels.

- New `xfuser/core/vsa_h3_attention.py` for tile packing, FlexAttention mask, and CPU tests
- USP all-to-all generalized so the compression gate travels with QKV
- MiniMax-H3 transformer wrapper loads `to_gate_compress` and runs VSA-H3 after RoPE when `enable_fasth3_vsa` is set; base MiniMax-H3 stays dense USP
- Runner `FastH3` / `FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-DataFree` registers T2VA with the 4-forward / 5-scheduler-point Preview v1 contract

## Important notes

- **T2VA only** for Preview v1. Other MiniMax-H3 tasks stay on the dense H3 runner.
- **Not Wan/Jenga VSA.** Different tile size, prefix handling, and sparsity pattern.
- **`--use_torch_compile` is rejected** on FastH3. FlexAttention is compiled independently.
- **`--use_hybrid_attn_schedule` is rejected.** VSA-H3 runs on every transformer step.
- Base MiniMax-H3 is unchanged unless FastH3 IDs / `enable_fasth3_vsa` are used.

## Test plan

- [ ] `pytest tests/core/test_vsa_h3_attention.py`
- [ ] `pytest tests/test_minimax_h3.py`
- [ ] FastH3 T2VA smoke run (5 scheduler points)
- [ ] Confirm MiniMax-H3 dense T2VA/I2VA still works without FastH3 flags

```bash
xdit --model FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-DataFree \
  --prompt "A red fox trots through a snowy pine forest." \
  --height 768 --width 1344 --num_frames 124 \
  --num_inference_steps 5 --seed 42 \
  --ulysses_degree 8 --attention_backend AITER --task t2va

### CODE REVIEW

File: xfuser/core/distributed/attention_backend.py [hunk 1]
```diff
@@ -1733,6 +1734,97 @@ def _aiter_attn_call(query, key, value, dropout_p, is_causal, attention_kwargs=N
...
+        return _aiter_attn_call(
+            query, key, value, dropout_p, is_causal, attention_kwargs
+        )
+    return _sdpa_attn_call(
+        query, key, value, dropout_p, is_causal, attention_kwargs
+    )
+
+
+@register_attention_function(AttentionBackendType.FLEX_VSA_H3)
+@torch.compiler.disable
```
avjves: Is compiler disable required? This causes a graph break, though maybe not noticeable in video models

xfuser/model_executor/models/transformers/transformer_minimax_h3.py
```diff
@@ -82,12 +115,17 @@ def __call__(
         }
         if use_ulysses:
             attention_args["combine_qkv_a2a"] = True
-        hidden_states = attention_function(
-            query.transpose(1, 2),
-            key.transpose(1, 2),
-            value.transpose(1, 2),
-            **attention_args,
-        ).transpose(1, 2)
+        try:
```
avjves: Why try-finally? Shouldn't it just crash if the attention_function fails?

xfuser/model_executor/models/transformers/transformer_minimax_h3.py
```diff
@@ -114,13 +208,26 @@ def __init__(self, *args, attention_backend=None, **kwargs) -> None:
...
                     backend=attention_backend,
+                    use_fasth3_vsa=enable_fasth3_vsa,
                 )
             )
 
         self.register_forward_pre_hook(
             lambda module, args: get_runtime_state().increment_step_counter()
         )
 
+    def _selected_attention_backend(self):
```
avjves: Why are these re-implemented here? Can you just share the functions with the attn processor calls?

avjves: Maybe simpler to set the attention backend in the __init__ so you can just call self.attention_backend and not do the checks and possibly reading from runtime state.

avjves (commented):
Thanks! Looks pretty good! I left a few comments :)

hakankiymaz-amd:
Thanks for the review, I have resolved those :) @avjves

avjves (approved):

PR #781, opened 2026-09-15
[Fix] Recompiles on Minimax-H3

nsakkine:
Fix recompiles in Minimax-H3 workload

#### Background

A Minimax-H3 bug was found by running benchmarks for multiple iterations and noticing that the first takes significantly more time than the rest. The reason was pinpointed to a recompile issue by running with `TORCH_LOGS=recompiles` and inspecting the logs.

The first timed iteration ran slower than the rest because a Dynamo recompilation landed inside it instead of during compile warmup. Two guard failures caused it.

-  `_usp_attention_kwargs` key set mutates between forwards. The wrapper `update()`s three varlen keys when the packed sequence needs padding and `pop()`s them when it does not. Dynamo guards on dict key membership, so the flip invalidates the guard.
- Timestep batch-size specialization. The timestep batch goes 1 → 2 → 1 across warmup and the timed iterations, and Dynamo recompiles on each transition.

#### Tasks

Fix the recompilation issues with

- [x] `transformer_minimax_h3.py`: keep all three keys always present, carrying `None` when there is no padding, so the key set is invariant. Behaviour is unchanged — every consumer already tests `.get(...) is None` rather than the dict's truthiness.
- [x] `minimax_h3.py`: `mark_unbacked` the timestep batch dimension before entering the compiled region. The wrapper uses `functools.update_wrapper` because the denoise block picks its layout kwargs `inspect.signature(transformer.forward).parameters`, and a bare `(*args, **kwargs)` wrapper would silently drop `token_tags`.

#### Tests

Fix tested by rerunning Minimax-H3 after the fixes had been deployed. The results

```
xdit --model MiniMaxAI/MiniMax-H3 --prompt 'A cat flying through the universe on a broom, playing rap music from his boombox' --height 768 --width 1344 --seed 42 --num_frames 124 --num_inference_steps 50 --num_iterations 3 --ulysses_degree 8 --use-parallel-vae --text_encoder_tp_degree 8 --use_torch_compile --attention_backend aiter
...
INFO 09-14 13:56:19 [runner_utils.py:30] Model initialization complete.
INFO 09-14 13:56:19 [runner_utils.py:30] Running model...
INFO 09-14 13:56:19 [runner_utils.py:30] Running iteration 1/3
INFO 09-14 13:57:22 [runner_utils.py:30] Iteration 1 completed in 27.17s
INFO 09-14 13:57:22 [runner_utils.py:30] Running iteration 2/3
INFO 09-14 13:58:24 [runner_utils.py:30] Iteration 2 completed in 27.14s
INFO 09-14 13:58:24 [runner_utils.py:30] Running iteration 3/3
INFO 09-14 13:59:31 [runner_utils.py:30] Iteration 3 completed in 27.13s
...
```

show that there is no substantial change between the relevant runtimes iteration-to-iteration. Furthermore, the logs were checked and found clean from recompile messages, unlike before the fix.

#### Other

Not tested on other attention backends.


xfuser/model_executor/models/runner_models/minimax_h3.py
```diff
@@ -537,14 +538,40 @@ def _compile_model(self, input_args: dict) -> None:
...
-        transformer.forward = torch.compile(
-            transformer.forward,
+        original_forward = transformer.forward
+        compiled_forward = torch.compile(
+            original_forward,
             mode=mode,
             fullgraph=not use_hybrid,
         )
+
+        # mark_unbacked is only exposed on torch._dynamo.decorators, not on
```
avjves, line 548: This section could use some background in comments. I at least csouldn't infer from this comment _why_ this is nessary 🤔

nsakkine: Thanks for being precise. True, it is not absolutely mandator as it just fixes one recompile during the compile-warmup pass. The point is that timestep is a 1D tensor whose shape changes along denoising (sometimes it contains distinct timesteps for video and audio, sometimes not) and this forces a recompile. Without this part of the recompile fix, there would be 2 graphs at actual execution from which the correct would be always selected, while with the fix, there is a single one with dynamic timestep-input. I have added some detail to comments, but should I remove this altogether?

avjves: If it affects perf, we should keep it. 
But could you add like simple comment before all of this e.g.:
``` 
The below fixes a recompile and therefore doesn't split the graph into two.
```

Something super simple that tells why it's there directly without having to infer :)

nsakkine: Simple clarifying comment there now 👍

avjves (approved):

PR #782, opened 2026-09-15
feat: add MXFP6 GEMM and unified GEMM configuration

ksikiric:
## Summary
- Add AITER A6W6-backed MXFP6 linear and loading support on gfx950.
- Support pure MXFP6 and mixed MXFP4+MXFP6 modes for Wan.
- Gemm CLI refactoring (old flags kept with deprecation warnings).
  - Unified precision selection under --gemm_quantization <SPEC>: FP8, FP4, FP6, INT8, FP4/FP8, or FP4/FP6.
  - Added --gemm_config <PATH> for advanced targets and explicit schedules; precision remains CLI-owned. 

## Testing
- Unit, loader, and multirank regression tests pass.
- Verify that legacy flags still work as intended. 
- Wan 2.2 T2V verified end-to-end on MI355X.
  - Results: 
    - mxfp4 + mxfp6 shows ~2% improvement over the mxfp4 + fp8 baseline for Wan2.2 T2V
    - To reproduce: `xdit --model Wan-AI/Wan2.2-T2V-A14B-Diffusers     --ulysses_degree 8     --prompt "a motorcycle cruising along a coastal highway"     --height 720     --width 1280     --num_frames 81     --num_inference_steps 40     --guidance-scale 4.0     --attention_backend aiter_fp8 --cross_attention_backend aiter --use_parallel_vae --use-torch-compile --gemm_quantization low=fp4,high=fp6`

Output video mxfp4 + fp8 (baseline):

https://github.com/user-attachments/assets/10f0aa9b-6e89-4c62-84de-5f843491efc6

Output video mxfp4 + mxfp6:

https://github.com/user-attachments/assets/98ec5172-a813-499a-8d23-4adfc46ea006


xfuser/config/args.py
```diff
@@ -137,6 +142,24 @@ def _normalize_determinism_check_report_ranks(value: str) -> frozenset[int]:
     return ranks
 
 
+def _add_gemm_profile_args(parser) -> None:
```
avjves, line 145: Why are these in a separate function? 🤔

xfuser/config/args.py
```diff
@@ -214,15 +237,24 @@ class xFuserArgs:
...
     cross_attention_backend: Optional[str] = None
+    gemm_quantization: Union[GemmQuantizationSpec, str, None] = None
+    gemm_config: Optional[str] = None
     use_int8_gemms: bool = False
     use_fp8_gemms: bool = False
     use_fp8_text_encoder: bool = False
     use_fp4_gemms: bool = False
+    # Alone, route the runner's FP4/FP8 target union to AITER MXFP6. With
+    # use_fp4_gemms, keep MXFP4 primary and use MXFP6 for its remainders.
+    use_fp6_gemms: bool = False
```
avjves, line 247: Do we want to add this? Seems like we should force the fp6 gemms through the new config as no-one is using them, yet.

xfuser/runner.py
```diff
@@ -74,7 +74,14 @@ def run(self, input_args: dict) -> Tuple[DiffusionOutput, list]:
 
     def preprocess_args(self, input_args: dict) -> dict:
         """ Preprocess input arguments before passing them to the model """
-        return self.model.preprocess_args(input_args)
+        resolved = dict(input_args)
+        for name in (
```
avjves: This seems pretty hacky 🤔 Any way we could do this without adding specific args here?

xfuser/model_executor/models/runner_models/base_model.py
```diff
@@ -1056,18 +1071,42 @@ def _setup_hybrid_attn_schedule(self, input_args: dict) -> None:
...
         multiplier = self._calculate_hybrid_attention_step_multiplier(input_args)
         total_steps = input_args["num_inference_steps"] * multiplier
-        num_high_precision_steps = input_args["num_hybrid_gemm_high_precision_steps"] * multiplier
 
-        gemm_schedule = create_hybrid_gemm_schedule(
-            num_high_precision_steps=num_high_precision_steps,
-            total_steps=total_steps,
-        )
+        if self.config.hybrid_gemm_schedule is not None:
+            from xfuser.core.distributed.attention_schedule import GemmPrecisionSchedule
```
avjves: nit: safe to import earlier?

xfuser/model_executor/models/runner_models/loading/format_backends.py
```diff
@@ -173,8 +181,52 @@ def _probe_aiter_mxfp4_apis() -> tuple[bool, str | None]:
     return _probe_aiter_fp4_kernels()
 
 
+def _probe_aiter_mxfp6_apis(
+    gcn_arch_probe: Callable[[], str | None] | None = None,
+) -> tuple[bool, str | None]:
+    """Probe the exact gfx950 AITER A6W6 surface used by xFuserMXFP6Linear."""
+
+    if os.getenv("AITER_TRITON_ONLY", "0") == "1":
```
avjves: Is this a real worry? I don't know so I'm asking 😄  Just wondering if this is something we need to check - what happens if one has that enabled? Is there a precedence to keep that on?

xfuser/core/utils/runner_utils.py
```diff
@@ -510,28 +560,54 @@ def quantize_linear_layers_to_fp4(
     use_hybrid_schedule: bool = False,
     device: Optional[torch.device] = None,
     filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
+    use_fp6_for_overrides: bool = False,
 ):
-    from torchao.quantization.granularity import PerTensor
-    from torchao.quantization.quant_api import Float8DynamicActivationFloat8WeightConfig, quantize_
     from xfuser.model_executor.layers.mxfp4_linear import xFuserMXFP4Linear, xFuserHybridMXFP4Linear
 
+    if not use_fp6_for_overrides:
```
avjves, line 567: Why this gated importing?

avjves: Also below

xfuser/model_executor/models/runner_models/loading/meta_load.py
```diff
@@ -588,7 +588,9 @@ def plan_text_encoders(self, existing_quantization_config=None):
 
     def materialize_pipeline(self) -> None:
         """Place or shard the loaded pipeline according to the current run config."""
-        if self.model.config.use_fp4_gemms:
+        if hasattr(self.model.config, "gemm_quantization_spec"):
```
avjves: Isn't that always true due to it being a property? Do we want to log the plan when we don't have a proper spec defined?

avjves: And the elif would never run?

avjves (commented):
Thanks!

One comment that was harder to leave as code lvl msg:
- --use_fp8_text_encoder is now unreachable on tiered configs. It expects use_fp8_gemms to be on, but low=fp4,high=fp8 does not set them. It should check the gemm quant spec instead. 


But overall this refactor goes towards the better direction, but it's still using a lot of use_fp6_gemms, use_fp8_gemms etc here and there and the quant functions have custom params specific to a precision, e.g. "fp6_precision_overrides" (Why not fp8 as well?). Maybe it's out of the scope of this PR, but this should be fairly high priority to change at some point. There's now too little abstraction - everything is very rigidly designed to just for current use cases and the next addition then has to inject it into this again.

We have at least these:
```
        fp8_gemm_module_list=["transformer.blocks"],
        fp8_text_encoder_module_list=["text_encoder.encoder.block"],
        fp4_gemm_module_list=["transformer.blocks"],
        fp8_precision_overrides=("0.", "1.", "2.", "3.", "4.",
                                 "5.", "6.", "7.", "8.", "9.",
                                 "30.", "31.", "32.", "33.", "34.",
                                 "35.", "36.", "37.", "38.", "39."),
```

We should refactor these to also _not_ be fp8/fp4 specific.


Anyway, I think that's beyond the scope of this PR :)


Btw, did you test the previous usage still works? The PR description only shows that you tested the mxfp6 one.

xfuser/config/args.py
ksikiric: Both add_cli_args and add_runner_args expose these options. The helper keeps their definitions consistent without duplicating them

xfuser/config/args.py
ksikiric: Yea true... I'll remove it and only the new --gemm_quantizaiton flag can enable fp6 gemms from now on

xfuser/model_executor/models/runner_models/loading/format_backends.py
ksikiric: Woops... this was something my agent added that I missed... this is only an issue if a user enables this flag, but I don't see why someone would? I'll remove this check

xfuser/core/utils/runner_utils.py
ksikiric: FP4/FP6 is an AITER-only path and does not use TorchAO. Since TorchAO is optional, importing it unconditionally could break a valid AITER-only configuration

ksikiric:
Thanks! I fixed --use_fp8_text_encoder to validate against the resolved GEMM spec, so tiered FP4/FP8 profiles now work. I agree the remaining format-specific internal fields should eventually become generic precision tiers, but I kept that outside this PR to limit scope. I can continue working on the GEMM refactoring on a separate PR after this has been merged. 

Previous behavior was also tested. The regression suite covered both new and legacy FP8/FP4/INT8 selectors, and the Wan2.2 FP4+FP8 baseline was run end-to-end. I’ll clarify that in the PR description.

avjves (approved):
Thanks, LGTM :) Merging

PR #783, opened 2026-09-16
Extended FP8 comms support

pds-amd:
- Generalise FP8 Ulysses all-to-all (`--use_fp8_comms`) beyond Wan.
- Quantizes Q/K/V to FP8 before the Ulysses collectives so the big host-staged all-to-alls move half the bytes. The dense FP8 attention kernel still runs; it now consumes those pre-quantized tensors with the calibrated descales. Still requires a one-shot calibration pass which we should investigate approaches to eliminate in future.

## Performance
- Paired runs on RDNA4 (no GPU P2P; collectives stage through host DRAM). Same image, NUMA policy, and `aiter_flydsl_fp8` backend; only `use_fp8_comms` flipped. Z-Image and SD3.5 have no 2-GPU row because that layout is CFG-parallel rather than Ulysses.
- all-to-all is host-bound here; on machines with P2P the speed up should be less.
- To my eye the outputs all look the same, but they are not bitwise identical.

| Workload | GPUs | use_fp8_comms: false (s) | use_fp8_comms: true (s) | Speedup |
|---|---:|---:|---:|---:|
| flux.usp_1k | 2 | 9.086 | 8.359 | +8.69% |
| flux.usp_1k | 4 | 7.019 | 5.665 | +23.89% |
| flux.usp_2k | 2 | 44.525 | 40.881 | +8.91% |
| flux.usp_2k | 4 | 25.152 | 22.619 | +11.20% |
| flux2.i2i_1k | 2 | 49.187 | 47.652 | +3.22% |
| flux2.i2i_1k | 4 | 34.307 | 33.353 | +2.86% |
| flux2.multi_1k | 2 | 61.152 | 56.959 | +7.36% |
| flux2.multi_1k | 4 | 41.631 | 39.626 | +5.06% |
| flux2.t2i_1k | 2 | 34.851 | 34.132 | +2.11% |
| flux2.t2i_1k | 4 | 28.413 | 27.947 | +1.67% |
| flux2_klein | 2 | 8.113 | 7.441 | +9.03% |
| flux2_klein | 4 | 4.395 | 3.944 | +11.45% |
| flux2_klein_4b | 2 | 4.524 | 4.096 | +10.43% |
| flux2_klein_4b | 4 | 2.532 | 2.298 | +10.18% |
| flux_kontext | 2 | 18.133 | 15.780 | +14.91% |
| flux_kontext | 4 | 13.151 | 11.063 | +18.88% |
| qwen_image | 2 | 131.273 | 119.723 | +9.65% |
| qwen_image | 4 | 88.834 | 82.086 | +8.22% |
| qwen_image_edit | 2 | 216.349 | 208.076 | +3.98% |
| qwen_image_edit | 4 | 137.357 | 126.725 | +8.39% |
| stablediffusion_3_5 | 4 | 4.700 | 4.056 | +15.87% |
| wan2_2_ti2v_5b.i2v | 2 | 136.093 | 129.768 | +4.87% |
| wan2_2_ti2v_5b.i2v | 4 | 75.302 | 70.038 | +7.52% |
| z_image | 4 | 25.186 | 23.850 | +5.60% |


xfuser/model_executor/layers/usp.py
```diff
@@ -366,20 +375,13 @@ def USP(
...
+            query, key, value = _combined_qkv_all_to_all(query, key, value)
         else:
             query = _ft_c_input_all_to_all(query)
             key = _ft_c_input_all_to_all(key)
             value = _ft_c_input_all_to_all(value)
-            for name, tensor in extra_inputs:
-                attention_kwargs[name] = _ft_c_input_all_to_all(tensor)
 
-    if attn_layer:
+    if attn_layer is not None and get_cache_manager().has_cache_entry(attn_layer):
```
avjves: Does this not break Pipefusion?

xfuser/model_executor/layers/usp.py
```diff
@@ -307,18 +298,43 @@ def USP(
...
-    joint paths.
+    joint paths. Also used as the FP8-comms module when ``attn_layer`` is None
+    (KV cache already updated by the caller).
     """
     if combine_qkv_a2a is None:
         combine_qkv_a2a = False
 
     attention_function = _get_attention_function(backend=backend)
 
+    fp8_module = attn_layer if attn_layer is not None else head_balance_layer
```
avjves, line 367: I'm a bit confused now. The changes to transformers now make it so that the `attn_layer` is passed always, but FP8 comms obj is not?

So auto_fp8 is always true?

avjves: Why is there auto_fp8 and then the fp8 comms object? 🤔

avjves (commented):
Thanks! Left a few comments, not sure if they require any changes.

xfuser/model_executor/pipelines/pipeline_stable_diffusion_3.py
```diff
@@ -439,13 +439,35 @@ def _init_sync_pipeline(self, latents: torch.Tensor, prompt_embeds: torch.Tensor
...
-                prompt_embeds = torch.chunk(prompt_embeds, get_sequence_parallel_world_size(), dim=-2)[get_sequence_parallel_rank()]
-            else:
-                get_runtime_state().split_text_embed_in_sp = False                
+            # pooled_prompt_embeds has no sequence dim; only the token stream is
+            # sharded. CFG already folded negative embeds into prompt_embeds.
+            prompt_embeds = self._pad_and_chunk_text_for_sp(prompt_embeds)
 
         return latents, prompt_embeds
 
+    def _pad_and_chunk_text_for_sp(self, prompt_embeds: torch.Tensor) -> torch.Tensor:
```
avjves: This now removes the joint attention - was this expected?

lauri9:
Regarding `bind_fp8_comms_attn_modules`, can you clarify how this interacts with hybrid attention schedules? We could have a mixed attention schedule with fp8 and bf16 (or another backend that is incompatible with fp8 comms).

xfuser/model_executor/layers/usp.py
pds-amd: Good catch. I updated USP to access the KV cache only when the supplied attention layer has a registered cache entry. I also added explicit validation rejecting the currently unsupported PipeFusion + FP8-comms combination, with tests for both cases.

xfuser/model_executor/layers/usp.py
pds-amd: Agreed, that split was confusing. I removed both auto_fp8 and the explicit per-call fp8_comms argument. USP now resolves FP8-comms state directly from the bound attention layer. Passing attn_layer does not itself enable FP8; it remains a no-op when FP8 comms is disabled or the layer has no synchronized scales.

xfuser/model_executor/pipelines/pipeline_stable_diffusion_3.py
pds-amd: Only for the FP8-comms execution path. Image and text Q/K/V are still concatenated, so the joint attention operation remains mathematically intact; it no longer uses the separate joint_strategy path because that cannot mix pre-quantized FP8 tensors with BF16 joint tensors. I updated the non-FP8 path to preserve the existing joint-attention behavior. For FP8, text is padded for the Ulysses collective and the synthetic trailing K/V entries are removed after gathering.

pds-amd:
> Regarding `bind_fp8_comms_attn_modules`, can you clarify how this interacts with hybrid attention schedules? We could have a mixed attention schedule with fp8 and bf16 (or another backend that is incompatible with fp8 comms).

@lauri9, good point to raise! The previous name was not great. I renamed it to `register_fp8_comms_eligible_modules`. Registration only identifies modules that can hold calibration state; USP still gates FP8 communications on the effective backend for each call. In a hybrid schedule, incompatible steps retain their normal communication path. I also added a test covering that behavior.

avjves (approved):
Thanks! LGTM now :)

PR #784, opened 2026-09-16
restore torch<2.10 compute-communication overlap passes

lauri9:
This PR restores the list of Inductor compute-communication overlap passes that were the default in `torch<2.10`.

## Background

In pytorch/pytorch#167080 (2005b5f, first in 2.10) the list `torch._inductor.config.reorder_for_compute_comm_overlap_passes` was emptied to `[]` in favor of `aten_distributed_optimizations`.

xDiT attempted in #696 to enable `aten_distributed_optimizations`, but this was reverted in #701 due to inductor bugs that could crash the process on newer PyTorch versions, especially 2.10.

The bugs have been fixed in upstream PyTorch, but 2.10 & co still being in active use it seems we might not want to fully reland #696.

The maintainers of PyTorch have [clarified that the old flags continue to work](https://github.com/pytorch/pytorch/issues/169461#issuecomment-3615822349) although they are no longer actively developed.

## Proposed change

Since `torch._inductor.config.reorder_for_compute_comm_overlap_passes` defaults to empty list in `torch>2.9`, currently xDiT silently disables the passes (and therefore compute-communication overlapping) for newer PyTorch versions. This PR reintroduces the default to not silently disable compute-communication overlapping when using `torch>2.9`.

## Performance impact
I experimented on several xDiT models on PyTorch 2.13, with and without this PR. The table provides a summary of runtimes on a MI300X platform. All configurations use Ulysses-8 sequence parallelism and bf16 attention; any other quantizations are explicitly mentioned.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/miklauri/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/miklauri/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
</head>

<body link="#467886" vlink="#96607D">

Model   configuration | This PR | Default flags | Improvement, %
-- | -- | -- | --
HunyuanVideo | 178.37 | 183.27 | 2.7%
HunyuanVideo, FP8 attn | 115.44 | 116.89 | 1.3%
Wan 2.1 | 134.62 | 135.97 | 1.0%
Wan 2.2 | 134.79 | 134.81 | 0.0%
Wan 2.2, FP8 GEMMs, FP8 attention | 77.28 | 78.61 | 1.7%

</body>

</html>

## Follow-ups

As a follow-up, the new `aten_distributed_optimizations` can be later enabled once they are stable and the performance is verified. They are intended to eventually replace `torch._inductor.config.reorder_for_compute_comm_overlap` and `torch._inductor.config.reorder_for_compute_comm_overlap_passes`


avjves (approved):
Thanks, LGTM!

PR #785, opened 2026-09-16
Add Lumina-Image-2.0 support with CFG and GQA-aware Ulysses

lavinal712:
## What does this PR do?

This PR adds xDiT acceleration support for Lumina-Image-2.0.

- Adds an `xFuserLumina2Pipeline` wrapper.
- Adds classifier-free-guidance parallelism with CFG degrees 1 and 2.
- Adds Ulysses sequence parallelism for Lumina2's joint text-image transformer blocks.
- Supports padded and variable-length text-image sequences through attention masks.
- Adds a GQA-aware USP path that keeps K/V heads compact during the Ulysses input all-to-all and repeats them locally after
communication.
- Adds public exports, runtime-state initialization, an example script, README support information, and focused USP tests.

## Motivation

Lumina2 uses grouped-query attention. Repeating K/V heads before the Ulysses all-to-all unnecessarily communicates the expanded K/V tensors.

The new USP path communicates compact K/V heads and performs KV repetition after all-to-all and KV-cache handling. This reduces communication volume without changing the generated output relative to the previous GQA implementation.

End-to-end validation used Lumina-Image-2.0 with BF16, 1024×768 output, 30 inference steps, guidance scale 4.0, CFG truncation ratio 0.25, and five seeds from 42 to 46 on two RTX 3090 GPUs.

| Setting | Mean latency | Speedup vs. 1 GPU |
|---|---:|---:|
| 1 GPU baseline | 12.609 s | 1.000× |
| 2 GPU CFG parallel | 10.343 s | 1.219× |
| 2 GPU Ulysses with legacy GQA communication | 11.714 s | 1.076× |
| 2 GPU Ulysses with compact GQA communication | 10.582 s | 1.192× |

## Current limitations

- Lumina2 sequence parallelism currently supports Ulysses only; ring_degree must be 1.
- The number of query and KV heads must be divisible by the Ulysses degree.
- PipeFusion and tensor parallelism are not included in this PR.


xfuser/model_executor/layers/usp.py
```diff
@@ -313,17 +389,49 @@ def USP(
...
     joint paths.
+
+    ``kv_head_repeat`` keeps grouped-query attention K/V heads compact during
+    the Ulysses input all-to-all, then repeats each local KV head immediately
+    before attention. With ``combine_qkv_a2a=True``, unequal Q and KV payloads
+    are packed into one collective.
     """
     if combine_qkv_a2a is None:
         combine_qkv_a2a = False
+    if isinstance(kv_head_repeat, bool) or not isinstance(kv_head_repeat, int):
```
avjves: This is just bunch of validation, right?
Can we keep the USP() function less cluttered by hoisting this to a validation function?
e.g.
```
if combine_qkv_a2a is None:
    combine_qkv_a2a = False

_validate_gqa_params(...)
```

avjves (commented):
Hi! Thanks for adding the support! 🙏

We have moved beyond the examples usage and use the model runner nowadays. If you want we can leave this as an example, but we are planning on deprecating the rest of the examples at some point as well, so this would be deprecated at the same time. Any chance you could add the support to the model runner instead? See for example how SD3.5 is implemented. See files `xfuser/model_executor/models/runner_models/stable_diffusion.py` and `docs/runner/runner.md`

xfuser/model_executor/layers/usp.py
lavinal712: Done. I moved the GQA argument checks into _validate_gqa_params()

lavinal712:
Thanks for the review! I have addressed both comments and rebased the PR onto the latest main.

- Added Lumina-Image-2.0 support to the unified model runner.
- Moved the GQA validation logic into _validate_gqa_params().

avjves (approved):
Looks good! Thanks for adding the support! 🙏

PR #788, opened 2026-09-17
feat: support CFG parallelism for Qwen Image Edit

pds-amd:
# Add CFG parallel support for Qwen Image Edit

## Summary

Enable CFG parallelism for Qwen Image Edit models, distributing the conditional and unconditional transformer branches across two ranks.

The implementation retains the upstream Diffusers pipeline flow while routing each rank's local prompt branch and gathering predictions before the scheduler step. Qwen's true-CFG normalization is preserved, and the existing Cache-DiT integration uses the CFG-aware cache mode now available on `main`.

## Changes

- Advertise `use_cfg_parallel` support for Qwen Image Edit.
- Add a Qwen Image Edit pipeline subclass that:
  - assigns the unconditional branch to CFG rank 0 and conditional branch to rank 1;
  - executes one transformer forward per rank and denoising step;
  - gathers both predictions and applies the original true-CFG combination and normalization;
  - falls back to the upstream pipeline when CFG parallelism is inactive.

## RDNA4 Performance

Qwen Image Edit at 2048×2048, 50 denoising steps, DBCache, five measured iterations:

| GPUs | Baseline topology | CFG topology | Baseline median | CFG median | Improvement |
|------|-------------------|--------------|-----------------|------------|-------------|
| 2 | Ulysses 2 | CFG 2 | 209.00 s | 168.74 s | 19.3% |
| 4 | Ulysses 4 | CFG 2 × Ulysses 2 | 151.92 s | 101.73 s | 33.0% |

## Validation

- Completed full 2-GPU and 4-GPU benchmark runs.
- Verified rank-local prompt routing and true-CFG prediction normalization.
- Verified Cache-DiT selects `enable_separate_cfg=False` under CFG parallelism.


avjves (approved):
LGTM, thanks! 🙏

PR #789, opened 2026-09-18
Add FastH3 V2 runner and V1 weight variant aliases

jjuvonen-amd:
Add `xFuserFastH3V2Model` and register two additional V1 weight checkpoints.

New FastVideo/FastH3 supported models:
- V2 8step [FastVideo/FastH3-V2](https://huggingface.co/FastVideo/FastVideo-FastH3-8-Step-V2)
- V1 4step weight variant [FastVideo/FastH3-Preview-v1-VSA-Synthetic-Step1300](https://huggingface.co/FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-Synthetic-Step1300)
- V1 4step weight variant [FastVideo/FastH3-Preview-v1-VSA-Synthetic-Step1900](https://huggingface.co/FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-Synthetic-Step1900)

### Changes

- Replace `FASTH3_MODEL_ID` with named constants `FASTH3_V1_DATAFREE_MODEL_ID` and
  `FASTH3_V2_MODEL_ID`; collect them in `FASTH3_MODEL_IDS` for iteration. Add
  `FASTH3_V1_MODEL_IDS` frozenset covering all three V1 checkpoints.
- Register `xFuserFastH3Model` under the Step1300 and Step1900 weight variants.
  Add `_customize_settings` override so `from_pretrained` receives the correct
  checkpoint path rather than always defaulting to the DataFree weights.
- Add `xFuserFastH3V2Model` (inherits V1 VSA-H3 backend; 9 scheduler points /
  8 transformer forwards; 768×1344×124 defaults).
- New supported models added to README.md and runner.md docs. 

### Tests
New models were confirmed to run on MI355X with good output video quality (videos exceed 10MB file size limit so they aren't attached).

**FastH3-V2**
```bash
xdit --model FastVideo/FastVideo-FastH3-8-Step-V2 \
  --task t2va \
  --prompt "A red fox trots through a snowy pine forest." \
  --height 768 --width 1344 --num_frames 124 \
  --num_inference_steps 9 --seed 42 \
  --ulysses_degree 8 \
  --attention_backend AITER \
  --use_torch_compile \
  --num_iterations 3 \
  --use_parallel_vae
```

**FastH3-V1 VSA-Synthetic-Step1900**
```bash
xdit --model FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-Synthetic-Step1900 \
  --task t2va \
[...]

PR #790, opened 2026-09-19
[Perf] MiniMax-H3: quantize decoded video on device instead of on the host

jjuvonen-amd:
## Problem

The MiniMax-H3 runner asks the pipeline for `output_type="np"`, which routes the decoded video through `VideoProcessor.postprocess_video` -> `pt_to_numpy`:

```python
images = images.cpu().permute(0, 2, 3, 1).float().numpy()
```

`.cpu()` transfers the whole float video to the host and `.permute()` only creates a view, which the `np.stack()` in `postprocess_video` then materializes. At the default 768x1344 / 124 frames that is a 1.43 GiB device-to-host copy followed by a 1.43 GiB NCHW→NHWC transpose running single-threaded in NumPy.

From a torch profiler trace of one T2VA run (MI355X, Ulysses degree 8):

```
+2073.0  Memcpy DtoH  1464.75 MiB    91 ms
+2164.2 .. +2329.1                  165 ms   no GPU work at all
+2329.2  audio decode starts
```

256 ms between the video decode and the audio decode, 165 ms of it with the GPU completely idle. The float result is then discarded: `encode_video_with_audio` takes uint8 `[F, H, W, C]`, and `_video_to_uint8` re-derives exactly that with `(x * 255).round().clip(0, 255)` on the host.

Trace of VAE decoding and audio decoding shows idle gap on GPU stream before changes:
before_decoding_gpu_idle

The gap is gone from trace after changes:
after_decoding_change

## Change

Ask for `output_type="pt"` and do the scale/round/clamp/cast and the transpose on the device, then copy. The transfer becomes 366 MiB already in the layout the encoder wants, and the host does no per-pixel work.

## Result

768x1344, 124 frames, one MI355X:

| | before | after |
|---|---|---|
| in `_run_pipe`, inside the timed region | 206.3 ms | 32.5 ms |
| in `save_output`, `_video_to_uint8` | 398.9 ms | passthrough |

So ~0.17 s off reported latency, plus ~0.4 s of wall clock that the latency number never included.

Encoded frames are unchanged. `_video_to_uint8` was already applying the same `(x * 255).round().clip(0, 255)`, so moving that arithmetic to the device is bit-identical. Verified at 768x1344x124 size, max abs difference 0.

## Notes

- `_decoded_videos_to_frames` returns a list so `DiffusionOutput` keeps unpacking batch items individually; a stacked tensor would be treated as a single output.
- `.float()` before scaling matches `pt_to_numpy`, which also casts before quantizing, so the result holds if the VAE emits a narrower float dtype.
- `_video_to_uint8` is left as is. It is a passthrough for the H3 path now but still covers the other input shapes `save_output` accepts.
- Both `_run_pipe` implementations are covered (T2VA/I2VA/L2VA/FL2VA and Ref2VA).

## Drive-by test fix

`tests/test_minimax_h3.py` currently fails to import on `main`: #789 renamed `FASTH3_MODEL_ID` to `FASTH3_MODEL_IDS` in the runner but left the test importing the old name, so `test_fasth3_defaults_match_inference_contract` and `test_fasth3_loads_published_checkpoint` error out. Since this PR touches that file it points them at `FASTH3_V1_DATAFREE_MODEL_ID`, which is what `xFuserFastH3Model.settings.model_name` is set from.

## Testing

`pytest tests/test_minimax_h3.py`: 2 failed / 35 passed before, 39 passed after. The Ref2VA fake-pipe stub moves to the `pt` layout and asserts the requested `output_type`, plus two new tests covering the conversion: contiguous uint8 `[F, H, W, C]` with correct channel values, and that an already-quantized video is only relaid out.


avjves (approved):
LGTM, thanks :)

PR #791, opened 2026-09-21
[Perf] LTX2.5: balance decode ranks using tile volume based cost proxy

lauri9:
## Problem

The LTX-2.5 diffusion decoder creates 12 overlapping tiles for the 121×1024×1536 output. Their feature-grid volumes are highly unequal:

| Tile sizes | Values |
|---|---|
| Temporal | 40, 33 |
| Height | 96, 40 |
| Width | 96, 96, 16 |

The current owner rule is `tile_idx % sp_world_size`. At sequence-parallel world size 8 this gives:

| Rank | Tile count | Assigned volume | vs. balanced share |
|---:|---:|---:|---:|
| 0 | 2 | 419,328 | 1.62× |
| 1 | 2 | **495,360** | **1.92×** |
| 2 | 2 | 188,160 | 0.73× |
| 3 | 2 | 174,720 | 0.68× |
| 4 | 1 | 153,600 | 0.60× |
| 5 | 1 | **25,600** | **0.10×** |
| 6 | 1 | 304,128 | 1.18× |
| 7 | 1 | 304,128 | 1.18× |

The collective assembly cannot progress faster than the slowest owner. Rank 1 therefore determines the compute critical path while rank 5 receives only 5.2% as much work. The analytical parallel efficiency is 52.11%, equivalent to only 4.17 fully utilized ranks.

## Proposed solution (This PR)

Assign tiles with deterministic longest-processing-time (LPT) scheduling:

1. Calculate each tile's feature-grid volume, `(t1-t0) * (h1-h0) * (w1-w0)`.
2. Sort tiles from largest to smallest, breaking ties by tile index.
3. Assign each tile to the currently least-loaded rank, breaking ties by rank.
4. Use the resulting owner map for both local decode and tile broadcasts.

Every rank computes the same map locally. Tile order, tile indices, per-tile seeds, decode operations, and final assembly remain unchanged.

At world size 8 the proposed assignment is:

| Rank | Tile count | Assigned volume | vs. balanced share |
|---:|---:|---:|---:|
| 0 | 1 | **368,640** | 1.43× |
| 1 | 1 | **368,640** | 1.43× |
| 2 | 1 | 304,128 | 1.18× |
| 3 | 1 | 304,128 | 1.18× |
| 4 | 2 | 179,200 | 0.69× |
| 5 | 2 | 174,720 | 0.68× |
| 6 | 2 | 188,160 | 0.73× |
| 7 | 2 | 177,408 | 0.69× |

The largest tile itself has volume 368,640, so no indivisible-tile assignment can have a lower critical path.

## GPU A/B validation

Tested on 8xMI300X platform, with Ulysses-8 config.

| Metric, median | Round robin (current main) | LPT (This PR) | Reduction |
|---|---:|---:|---:|
| Slowest rank's local tile compute | 4,734.6 ms | 3,782.3 ms | **20.1%** |
| Tile-parallel section | 4,831.2 ms | 3,839.8 ms | **20.5%** |
| End-to-end iteration | 10,996.6 ms | 9,779.4 ms | **11.1%** |

We see that LPT balancing leads to clear reductions in median times for the slowest rank local tile compute and end-to-end runtimes.

## Caveats / possible follow-ups
- GPU validation covers SP=8 at 121×1024×1536. Other geometries are analytically improved but not benchmarked.
- Tile volume is a cost proxy; aspect ratio, kernel overhead, or other such factors can make runtime non-linear.
- LPT cannot overcome an oversized indivisible tile; tile geometry still limits maximum efficiency. It is also not guaranteed to be globally optimal, but seems to be clear improvement over round-robin assignment


avjves (approved):
Nice, LGTM!

PR #792, opened 2026-09-22
Enable FP8 communications with AITER MHA v4

ksikiric:
## Summary
- route FP8-communicated, pre-quantized Q/K/V directly through the eligible AITER MHA v4 packed kernel
- pass the existing per-tensor descales and native FP8 format without requantizing or rotating again
- retain the legacy dense FP8 attention path as the fallback

## Scope
This change affects only the dense `AITER_FP8` MHA_v4 backend.

## Test
In a single 8-GPU Wan 2.2 T2V run, running MHA_v4 with FP8 comms reduced E2E latency with 5.6% while preserving output quality.


avjves (approved):
LGTM :)

